PR #1, opened 2016-10-15
Tune default configuration to be more friendly

kalessil:
By default, the plugin reports all empty and isset usages, so new users in 90% of cases got frustrated.

This PR deactivates the behaviour but leaves hints regarding null comparison, count usage and etc.


kalessil:
@ocramius ,  @samdark , @dryabov , @nazar-pc, @funivan : would appreciate your thoughts on this PR, perhaps there are more inspections which frustrate.

nazar-pc:
I've already disabled them in my project, so :+1: from me.

funivan:
It will be OK if this inspection will be disabled by default.

samdark:
I support disabling it by default.

What I have turned off (not sure it's default or not):

- Empty class
- Unnecessary last comma in array definition
- Usage of inclusion return value

The rest is turned on.

funivan:
In my projects I enable `Suggests to use null comparison` and disable `report empty() usage` and `Suggest to use count() - comparions`
I think this is more about code style than code performance. 
```php
$user = Db::fetchById(1);
if($user!==null){}  // easy to read. 
if(!empty($user)){} // hard to read

$users = Db::getAllByIds([1,2,3]);
if(!empty($users)){} // easy to read
if(count($users)>0){} // hard to read
```
`Isset` inspection is disabled. Easy to read and after some years of coding I know difference between `isset` and `array_keys_exists`
This is only IMHO ;)

kalessil:
@funivan : count usage hint will be disabled by default, other things you mentioned already here.

kalessil:
@samdark : first two deactivated by default, so I wondering about "Usage of inclusion return value".
Could you please share details why it's not usable for you?

samdark:
@kalessil because it is how I store/use configs. The config itself is:

```php
<?php
return [
    'key' => $value,
];
```

Usage is:

```php
$config = require 'config.php';
(new Application($config))->run();
```

funivan:
@kalessil from the symfony https://github.com/symfony/symfony/blob/5129c4cf7e294b1a5ea30d6fec6e89b75396dcd2/src/Symfony/Component/Translation/Loader/PhpFileLoader.php#L26

kalessil:
I see and it seems that it is already disabled by default, so we on the same page now.

PR #76, opened 2016-12-13
proper use of type API

neuro159:
that method was not needed and is dropped in 2017.


kalessil:
Thank you @neuro159 , looks even more logical than with utils.

PR #87, opened 2016-12-28
Add plugin badges

Haehnchen:


kalessil:
Looks great @Haehnchen, thanks for an awesome service)
Question about the last widget (downloads last month) - it shows n/a, is it correct?

Haehnchen:
yes "n/a" is correct for now. "last month stats" were delivered by JetBrains api before, now this is non public. so this calculation is done over montly diff, it needs atleast 2 month of data before showing valid values.

PR #88, opened 2017-01-02
Standardize inspection messages

tomzx:
- Ensure ending dot.
- Replace use of "shall" with "should" to better respect RFC2119.
- Attempt to normalize function call identification (use of 'funcname()' in messages).

A few additional notes:
- It appears that some tests are currently failing (without these changes)
- `com.kalessil.phpStorm.phpInspectionsEA.security.CryptographicallySecureAlgorithmsInspectorTest` is missing its fixture file


src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/phpDoc/UnknownInspectionInspector.java
```diff
@@ -77,8 +79,7 @@ public void visitPhpDocTag(@NotNull PhpDocTag tag) {
                 if (reported.size() > 0) {
                     final PsiElement target = tag.getFirstChild();
                     if (null != target) {
-                        final String message = "Unknown inspection: %i%".replace("%i%", String.join(", ", reported));
-                        holder.registerProblem(target, message, ProblemHighlightType.WEAK_WARNING);
+                        holder.registerProblem(target, message.replace("%i%", String.join(", ", reported)), ProblemHighlightType.WEAK_WARNING);
```
kalessil: It's better to keep `final String message` here - easier to read.

kalessil:
That is a huge work, huge thanks for fixing my English.

PS: true, UT is broken due to not provided test-cases.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/arrays/LowPerformanceArrayUniqueUsageInspector.java
```diff
@@ -12,8 +12,8 @@
 import org.jetbrains.annotations.NotNull;
 
 public class LowPerformanceArrayUniqueUsageInspector extends BasePhpInspection {
-    private static final String strProblemUseArrayKeysWithCountValues = "array_keys(array_count_values(<expression>)) will be more performing (but provide comments)";
-    private static final String strProblemUseCountWithCountValues     = "count(array_count_values(<expression>)) will be more performing (but provide comments)";
+    private static final String strProblemUseArrayKeysWithCountValues = "array_keys(array_count_values(...)) would be more efficient.";
```
tomzx: I'm not entirely sure it should be the responsibility of PHPInspectionEA to tell developers to add comments when they do these types of changes, which is the reason why I removed that part of the message. I think it should be left at the developer's discretion whether or not a change should receive a comment. I don't mind adding it back, however I'd definitely like to understand your rationale as to why in this case and not any of the other inspections.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/phpDoc/UnknownInspectionInspector.java
tomzx: Are you ok with `final String messageFormatted` as message is already part of the class now (`private static final String message = "Unknown inspection: %i%.";`)?

tomzx:
Left you two comments/questions. I'll wait for your feedback before applying the necessary changes. Do you prefer I add newer commits or that I amend this commit?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/arrays/LowPerformanceArrayUniqueUsageInspector.java
kalessil: The point is that we're suggesting to make code less readable here, so fair would be to put some obligations to the reader. Then, the reader will think a little bit if it's really worth it.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/phpDoc/UnknownInspectionInspector.java
kalessil: Oops, false-remark from me. Keeping as it is now ;)

kalessil:
Just push new commit, so PR got updated and we have a track of resolved remarks.

tomzx:
Alright, I've updated the message as discussed above.

Let me know if you see anything else that needs to be changed.

kalessil:
Looks very good, merging. Thank you @tomzx !

Issue #31, opened 2016-11-03
[followup] disallow writing into static properties (optional)


kalessil:
Original link: https://bitbucket.org/kalessil/phpinspectionsea/issues/299/disallow-writing-to-static-properties
Author @bram123

See original ticket

kalessil:
Shifting fresh-look review to 2.3.7, need to focus on 2.3.6

rentalhost:
I don't know...

If I can't write on static properties, then maybe I should use constants, instead this. I rarely create static properties. Basically I use to store a single instance, but it is an anti-pattern (now I am using the `App::singleton()` of Laravel to do that).

PR #150, opened 2017-02-21
#31 Disallow writing into static properties

funivan:
Here is prototype of the inspection. This is my vision of how it should work. If anyone have suggestions do not hesitate to write a comment. 
Some steps that I want to do before merge:
- Write tests
- Improve message and description
- Decide if this inspection should produce Warning by default.

### DIFF
--- a/META-INF/plugin.xml
+++ b/META-INF/plugin.xml
@@ -584,6 +584,11 @@ Bug-fix release, fixed a new bunch of false-positives<br/>
         shortName="ParameterDefaultValueIsNotNullInspection"      displayName="Non-null parameters default value"
         groupName="Code Style"                                    enabledByDefault="false" level="WEAK WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.ParameterDefaultValueIsNotNullInspector"/>
+    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
+        shortName="DisallowWritingIntoStaticPropertiesInspection" displayName="Disallow writing into static properties"
+        groupName="Code Style" enabledByDefault="true" level="WARNING"
+        implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.DisallowWritingIntoStaticPropertiesInspector"/>
+
 
 
     <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
@@ -694,12 +699,6 @@ Bug-fix release, fixed a new bunch of false-positives<br/>
         shortName="FopenBinaryUnsafeUsageInspection"                displayName="Binary-unsafe 'fopen(...)' usage"
         groupName="Compatibility"                                   enabledByDefault="true" level="WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.apiUsage.fileSystem.FopenBinaryUnsafeUsageInspector"/>
-
-
-    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
-        shortName="DisallowWritingIntoStaticPropertiesInspection" displayName="Disallow writing into static properties"
-        groupName="Code Style" enabledByDefault="true" level="WARNING"
-        implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.DisallowWritingIntoStaticPropertiesInspector"/>
 
 
 
--- a/META-INF/plugin.xml
+++ b/META-INF/plugin.xml
@@ -586,7 +586,7 @@ Bug-fix release, fixed a new bunch of false-positives<br/>
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.ParameterDefaultValueIsNotNullInspector"/>
     <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
         shortName="DisallowWritingIntoStaticPropertiesInspection" displayName="Disallow writing into static properties"
-        groupName="Code Style" enabledByDefault="true" level="WARNING"
+        groupName="Code Style" enabledByDefault="false" level="WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.DisallowWritingIntoStaticPropertiesInspector"/>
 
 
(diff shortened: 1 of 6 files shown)


META-INF/plugin.xml
```diff
@@ -695,6 +695,14 @@ Bug-fix release, fixed a new bunch of false-positives<br/>
         groupName="Compatibility"                                   enabledByDefault="true" level="WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.apiUsage.fileSystem.FopenBinaryUnsafeUsageInspector"/>
 
+
+    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
```
kalessil: Please group with other CS checks [here](https://github.com/funivan/phpinspectionsea/blob/8ab81cbc8f2832d325a9542e51ef5b7842720603/META-INF/plugin.xml#L583)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/DisallowWritingIntoStaticPropertiesInspector.java
```diff
@@ -0,0 +1,107 @@
...
+        return new BasePhpElementVisitor() {
+
+            @Override
+            public void visitPhpAssignmentExpression(AssignmentExpression assignmentExpression) {
+                PhpPsiElement variable = assignmentExpression.getVariable();
+                if (!(variable instanceof FieldReference)) {
+                    return;
+                }
+
+                if (!((FieldReference) variable).getReferenceType().isStatic()) {
```
kalessil: And not a constant, this would be invalid but we still don't want to process constant references.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/DisallowWritingIntoStaticPropertiesInspector.java
```diff
@@ -0,0 +1,107 @@
...
+                if (containingClass == null || callyClassFqn == null || !callyClassFqn.equals(containingClass.getFQN())) {
+                    // Property is modified in the other class method
+                    holder.registerProblem(assignmentExpression, message, ProblemHighlightType.WEAK_WARNING);
+                    return;
+                }
+
+
+                // Special case for inheritance. Check if property is declared in the cally class
+                String propertyName = variable.getName();
+                if (!isPropertyDefinedInClass(callyClassFqn, propertyName, holder.getProject())) {
```
kalessil: In this case it'll be more performing to make use of PS resolving APIs (pseudocode):
- FieldReference.resolve() instanceof Field
- field.getContainingClass().getFQN.equals(callyClassFqn )

Note: PhpIndexUtil belongs to "old" parts of the plugin, and I stopped using it in new checks completely.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/DisallowWritingIntoStaticPropertiesInspector.java
funivan: At line 38 we catch only `Field References` 
Constants does not belong to Field. They have  `Class constant reference`

kalessil: +1

META-INF/plugin.xml
funivan: Done

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/DisallowWritingIntoStaticPropertiesInspector.java
funivan: Done.

bram123:
This inspection has the warning text:
"Static property should be modified only inside the source class"

But the FR is about writing to static properties, at all.
Writing to properties from outside the owner's class is a separate inspection, and I'm thinking that's not only for static properties. I rather have all modifications go through a setter for possible validation.

funivan:
@bram123 So why then you should use static properties why you can not write to them? In this case it would be better to create inspection "Disallow static properties" But I thinks this is not a good idea.

bram123:
If you've got a big array as class property, that does not change. An inspection will trigger saying it's better to place it as a static property. So it's read only.

The problem with writing to static properties is that it can create problems with unit tests (you want a clean state at the start of every test, without having to cleanup all changes made). But also regular requests can have problems with this.

That's why we have the coding rule internally to disallow/discourage writing to static properties. And that's what triggered this feature request.

We don't want to disallow static properties all together, because they still have their uses. But only writing to them from any class/object, and unrelated also writing to properties from outside of the class.

funivan:
@bram123 you can use class constant instead

bram123:
@funivan Unfortunately, we can't yet because of PHP version.

Also, that would still make this a valid inspection, discouraging the use of writing to static properties because of reasons stated above.

funivan:
@bram123 i think we can solve this by adding option (Checkbox).
`[ ] Allow write to the static property from the source class` This checkbox will be checked by default.
If anyone want to disallow writing from any class then he can uncheck it and inspection will produce all Warnings on property modifications.

@kalessil what do you think ?

kalessil:
+1 for the option, enabled by default.

funivan:
Ok I will implement it as soon as possible)

funivan:
@kalessil waiting for approval ;)

kalessil:
Cool, will merge it as launched the indeegogo campaign.

kalessil:
Looks good to me.

Fleshgrinder:
We could check the configured PHP version for the project and recommend the usage of a `[public] const` instead of a `public static`.

funivan:
@Fleshgrinder only if option `Allow write from the source class` is unchecked ;)

META-INF/plugin.xml
```diff
@@ -584,6 +584,11 @@ Bug-fix release, fixed a new bunch of false-positives<br/>
         shortName="ParameterDefaultValueIsNotNullInspection"      displayName="Non-null parameters default value"
         groupName="Code Style"                                    enabledByDefault="false" level="WEAK WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.ParameterDefaultValueIsNotNullInspector"/>
+    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
+        shortName="DisallowWritingIntoStaticPropertiesInspection" displayName="Disallow writing into static properties"
+        groupName="Code Style" enabledByDefault="true" level="WARNING"
```
kalessil: @funivan: can we release it as optional (`enabledByDefault="false"`), later we can rename it and activate by default. I'd merge the PR immediately then/

The background is to make bug-fix and QF release - to support the crowdfunding campaign PR.

funivan: Done =)

kalessil:
Merging. @funivan @Fleshgrinder: would you please create a followup ticket if `[public] const` should happen?

PR #194, opened 2017-03-23
Add .gitignore Remove idea/workspace.xml from the git.

funivan:
@kalessil you accidentally add workspace.xml =) I try to fix this problem


kalessil:
Thank you @funivan !

Issue #183, opened 2017-03-13
::class can be used fix creates the use statement as a trait instead of an import if there's already a trait defined


Garethp:
When trying to use the fix for `::class can be used` on a property, and the class already has `use` statements for traits, the class is imported as a trait instead of a class import.

Steps to reproduce:
Have the following files

```php
namespace NamespaceOne;

class A
{
}
```
```php
namespace NamespaceTwo;

class B
{
    use C;
    
    protected $property = 'NamespaceOne\A';
}
```
```php
namespace NamespaceTwo;

trait C
{
}
```

Try to run the fix on `NamespaceTwo\B::$property` and see that `NamespaceOne\A` gets imported as a trait

kalessil:
Note for myself: check com.jetbrains.php.refactoring.PhpAliasImporter#insertUseStatement* methods

kalessil:
@Garethp : the fix works, thanks! I extended tests with testing fixes: 2593558f65d2cbdade5e6c2b70099388cc0e2214

Garethp:
Thanks. Are there any other quick fix bugs you'd like me to look at? Or test existing ones? I'd love to contribute, and those seem like good starting points for learning the plugin.

Also, are you able to explain the checkHighlighting function at all? I don't understand what the three booleans do, but any other values caused errors, and I wasn't able to get the available quick fixes without that function first.

kalessil:
Perhaps increasing test coverage will be a good point: https://github.com/kalessil/phpinspectionsea/blob/master/RULES.md#inspections-lists-type-compatibility: IsEmptyFunctionUsageInspection doesn't have tests yet. Best will be to reference to last commits for adding tests.

I don't remember the meaning of the flag, but this combination will apply given inspector, ensure that all warnings (error, warning, weak_warning "tags" in fixtures) and will fail when results are not as expected.

If the idea with test works, I'll assign a ticket for you.

PR #202, opened 2017-03-24
Adding a check to make sure that the use statement isn't a class trait

Garethp:
Fixes Issue #183 that I reported a week and a half ago

I'd love to contribute more, ~~~maybe write a test for this, but I couldn't run any of the tests. I got a weird message about not being able to find bundle messages.CssBundle when I tried to run the tests? I ended up figuring out how this worked and debugging by copying the popupBox you use for updates and using the title for debugging information~~~

Edit: I didn't follow your docs for setting up a dev environment (didn't even know they were there, thanks), so I didn't end up adding the CSS lib from PHPStorm, just the PHP and PHPStorm Lib (Because they were already in the settings, just pointing to a bad path)

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ClassConstantCanBeUsedInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ClassConstantCanBeUsedInspector.java
@@ -172,7 +172,8 @@ public class ClassConstantCanBeUsedInspector extends BasePhpInspection {
                     /* check all use-statements and use imported name for QF */
                     for (PhpUseList use : PsiTreeUtil.findChildrenOfType(file, PhpUseList.class)) {
                         /* do not process `function() use () {}` constructs or class traits */
-                        if (use.getParent() instanceof Function || use.getParent() instanceof PhpClass) {
+                        PsiElement parent = use.getParent();
+                        if (parent instanceof Function || parent instanceof PhpClass) {
                             continue;
                         }
 
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ClassConstantCanBeUsedInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ClassConstantCanBeUsedInspector.java
@@ -171,8 +171,8 @@ public class ClassConstantCanBeUsedInspector extends BasePhpInspection {
 
                     /* check all use-statements and use imported name for QF */
                     for (PhpUseList use : PsiTreeUtil.findChildrenOfType(file, PhpUseList.class)) {
-                        /* do not process `function() use () {}` constructs */
-                        if (use.getParent() instanceof Function) {
+                        /* do not process `function() use () {}` constructs or class traits */
+                        if (use.getParent() instanceof Function || use.getParent() instanceof PhpClass) {
                             continue;
                         }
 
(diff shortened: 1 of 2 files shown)


Garethp:
I'm trying to write a test for this quick fix, and so far I think I've been able to make the quickFix be applied, but it seems whenever I try to run `checkResultByFile` or whatever, it says the files don't exist. I'm guessing they don't get copied to the sandbox or something? Any ideas on how to address this? Having unit tests on the quick fixes would make it a lot easier to debug them by just setting a debug point in the LocalFix and executing it, you know?

kalessil:
That's great news @Garethp !

Probably tests configuration (see the selected option) needs adjustments:
image

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ClassConstantCanBeUsedInspector.java
```diff
@@ -171,8 +171,8 @@ public void applyFix(@NotNull Project project, @NotNull ProblemDescriptor descri
 
                     /* check all use-statements and use imported name for QF */
                     for (PhpUseList use : PsiTreeUtil.findChildrenOfType(file, PhpUseList.class)) {
-                        /* do not process `function() use () {}` constructs */
-                        if (use.getParent() instanceof Function) {
+                        /* do not process `function() use () {}` constructs or class traits */
+                        if (use.getParent() instanceof Function || use.getParent() instanceof PhpClass) {
```
kalessil: A local variable needed here: `final PsiElement parent = use.getParent();`

Garethp: Sorry, but my experience in Java is very limited. Can you explain why a local variable is needed?

kalessil: That;s the same as in PHP: duplicate function call. Instead, we just call it once and then do instanceof of checks against the parent variable.

Garethp: Ah, gotcha. I just assumed that Psi would cache the element after the first call. I'll introduce the change in a bit

Garethp:
My working directory is that, so I tried putting the file directly in that directory and it still didn't work. For reference, here's the code I'm trying to add:

```java
```

And here's the error I'm getting:

```
java.lang.RuntimeException: java.io.FileNotFoundException: /home/gareth/.IntelliJIdea2017.1/system/plugins-sandbox/test/class-name-constant-ns-fixed.php (No such file or directory)
```

kalessil:
In this case, we'll need to split the case into 2 different:
```php
    public void testIfFindsSameNsPatterns() {
        myFixture.configureByFile("fixtures/lang/classConstant/class-in-the-same-namespace.php");
        myFixture.enableInspections(ClassConstantCanBeUsedInspector.class);
        myFixture.testHighlighting(true, false, false);
    }

    public void testIfFindsAllPattern() {
        myFixture.configureByFile("fixtures/lang/classConstant/class-name-constant-ns.php");
        myFixture.enableInspections(ClassConstantCanBeUsedInspector.class);
        myFixture.testHighlighting(true, false, false);

        for (IntentionAction fix: myFixture.getAllQuickFixes()) {
            myFixture.launchAction(fix);
        }
        myFixture.checkResultByFile("fixtures/lang/classConstant/class-name-constant-ns.fixed.php"); // Please note '.fixed.php' naming here
    }
```

kalessil:
Just updated the last comment, let me know if the adjustment did help. If not, I'll take care about fixtures.

Garethp:
That didn't help, but when I tried this:

```java
        myFixture.setTestDataPath("fixtures/");
        myFixture.checkResultByFile("lang/classConstant/class-name-constant-ns.php");
```

I at least got 

```
com.intellij.rt.execution.junit.FileComparisonFailure: lang/classConstant/class-name-constant-ns.php 
```

So, from there I can actually write a corrected file to test against. I suppose for checking against files, you need to set the testDataDirectory. I'll add some tests tomorrow

Garethp:
For now I've just commited the variable extraction and the test, without a fix file. I made the test data directory the same as the working directory so that the paths for fixtures stayed consistent.

kalessil:
Thank you @Garethp!

PR #209, opened 2017-03-27
Multiple return points

funivan:


src/test/resources/fixtures/codeStyle/multiple-return-points.php
```diff
@@ -0,0 +1,54 @@
...
+            return false;
+        }
+
+        return true;
+    }
+}
+
+class Test_B
+{
+    public function customMethod()
```
kalessil: What about this example?

```php
private function ...() {
    return function(){ return ''; };
}
```

src/main/resources/inspectionDescriptions/MultipleReturnPointsInspection.html
```diff
@@ -0,0 +1,10 @@
+<html>
+<body>
+Reports methods with too many return points.<br>
+Methods with too many return points may be confusing, and hard to refactor.
+<br />
+<br />
+Please reference to corresponding <a href="https://stackoverflow.com/questions/36707/should-a-function-have-only-one-return-statement">stackoverflow thread</a>
+for more details.
```
kalessil: 👍

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MultipleReturnPointsInspector.java
```diff
@@ -0,0 +1,85 @@
...
+
+    @NotNull
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, boolean isOnTheFly) {
+        return new PhpElementVisitor() {
+
+            public void visitPhpMethod(Method method) {
+                final PhpClass containingClass = method.getContainingClass();
+                if (containingClass != null && !containingClass.isTrait()) {
+
+                    ReturnPointCountVisitor countVisitor = new ReturnPointCountVisitor(method);
```
kalessil: `final ReturnPointCountVisitor`

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MultipleReturnPointsInspector.java
```diff
@@ -0,0 +1,85 @@
...
+                    method.accept(countVisitor);
+
+                    if (countVisitor.getCount() > 1) {
+                        PsiElement nameIdentifier = method.getNameIdentifier();
```
kalessil: I'd suggest moving this into very beginning of the method

src/test/resources/fixtures/codeStyle/multiple-return-points.php
funivan: You mean method without name ?

kalessil: Almost, a method returning a lambda with a single return statement.
Returns in lambdas should not be counted.

kalessil: Thanks! Dropping the comment.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MultipleReturnPointsInspector.java
```diff
@@ -0,0 +1,83 @@
...
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, boolean isOnTheFly) {
+        return new PhpElementVisitor() {
+
+            public void visitPhpMethod(Method method) {
+                final PhpClass containingClass = method.getContainingClass();
+                final PsiElement nameIdentifier = method.getNameIdentifier();
+                if (containingClass != null && nameIdentifier != null && !containingClass.isTrait()) {
+
+                    final ReturnPointCountVisitor countVisitor = new ReturnPointCountVisitor(method);
+                    method.accept(countVisitor);
```
kalessil: @artspb, requesting for support =) 

What are advantages of using visitors vs PsiTreeUtil search methods+regular foreach loop? Are there any, and why if any?

artspb: `PsiTreeUtil` is good for traversing a tree upwards (to the root) or when you need to work with element's children. When you need to traverse a tree downwards (to the leafs) and you don't know an exact position of an element you're looking for or you need to work with several elements, a recursive element visitor is the best choice. It uses a "regular foreach loop" under the hood, but hides this irrelevant detail which makes the code more descriptive and allows concentrating on what you really want to do.

Working with PHP you usually use `PhpElementVisitor`. There's a recursive version of it called `PhpRecursiveElementVisitor`. It's marked as deprecated to recall you, that you should use it really carefully. When a tree is huge you can easily get a SOE or consume a lot of CPU for nothing. To avoid it a tree traversing should be stopped as soon as the desired element is found. For instance, the current implementation contains this mistake and will check each and every element for `instanceof PhpReturn`. A tree traversing won't be stopped even when more than 1 return point is found. That's why we usually don't recommend to use it at all.

**tl;dr**
1. Use `PsiTreeUtil` to find a parent or nearest children.
2. Use `PhpElementVisitor` to make your code more descriptive.
3. Never ever use `PhpRecursiveElementVisitor` (or any handmade recursive visitor) until you understand what you do. Don't forget to profile your code on really deep trees.

kalessil: That a great explanation, thank you!

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MultipleReturnPointsInspector.java
```diff
@@ -0,0 +1,83 @@
...
+    @NotNull
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, boolean isOnTheFly) {
+        return new PhpElementVisitor() {
+
+            public void visitPhpMethod(Method method) {
+                final PhpClass containingClass = method.getContainingClass();
+                final PsiElement nameIdentifier = method.getNameIdentifier();
+                if (containingClass != null && nameIdentifier != null && !containingClass.isTrait()) {
+
+                    final ReturnPointCountVisitor countVisitor = new ReturnPointCountVisitor(method);
```
artspb: As I mentioned, that a recursive visitor must be avoided, I should provide an alternative. In the context of this task, I'd recommend using a control flow. It can be easily obtained calling `method.getControlFlow()`. A building of a control flow is a kinda expensive task, but once built it's cached and can be reused by several inspections. So most probably some other inspection is already built it for you and you'll get it for free. (Disclaimer: never trust such sentences and always re-check using a profiler.)

Here's the sketch how the code can look like.

    final PhpControlFlow controlFlow = method.getControlFlow();
    final PhpExitPointInstruction exitPoint = controlFlow.getExitPoint();
    int count = 0;
    for (final PhpInstruction instruction : exitPoint.getPredecessors()) {
      if (instruction instanceof PhpReturnInstruction) {
        count++;
      }
      if (count > 1) {
        holder.registerProblem(...);
        return;
      }
    }

Most probably this code doesn't cover some corner cases like `finally`. So you should check how does it work and adjust accordingly. The `PhpControlFlowUtil` class can be helpful. (Note: it's not a part of the open API thus a subject to change.)

kalessil: That's really neat, how long the control flow stuff being there?

artspb: Since forever :)

kalessil: Oh my, I have to learn again. Both funny and annoying %)

kalessil: Thank you @artspb !!!

funivan: @kalessil should I rewrite code with using control flow technique ?

kalessil: @funivan : yes please, I'll take care about profiling the final version.

funivan: @kalessil done =)

kalessil: 👍, add please also @artspb in co-authors please =)

kalessil:
Thank @funivan! Evaluating performance topic this week.

PR #214, opened 2017-03-28
Adding a test for IsEmptyUsage Inspector

Garethp:
I've added a test, and in this test I've discovered a bug (That I'm going to make a shot at fixing). I'm not sure if it's a Php Inspections Bug, but the warning produced for `isIntNull` is wrong. It doesn't detect `$int` as an integer or a nullable for some reason, or it gives the weak warning of `empty` allowing too many things through.

### DIFF
--- a/src/test/java/com/kalessil/phpStorm/phpInspectionsEA/api/IsEmptyFunctionUsageInspectorTest.java
+++ b/src/test/java/com/kalessil/phpStorm/phpInspectionsEA/api/IsEmptyFunctionUsageInspectorTest.java
@@ -1,0 +1,17 @@
+package com.kalessil.phpStorm.phpInspectionsEA.api;
+
+import com.intellij.testFramework.fixtures.CodeInsightFixtureTestCase;
+import com.kalessil.phpStorm.phpInspectionsEA.inspectors.apiUsage.IsEmptyFunctionUsageInspector;
+
+final public class IsEmptyFunctionUsageInspectorTest extends CodeInsightFixtureTestCase {
+    public void testIfFindsAllPatterns() {
+        IsEmptyFunctionUsageInspector inspector = new IsEmptyFunctionUsageInspector();
+        inspector.SUGGEST_TO_USE_COUNT_CHECK = true;
+        inspector.REPORT_EMPTY_USAGE = true;
+        inspector.SUGGEST_TO_USE_NULL_COMPARISON = true;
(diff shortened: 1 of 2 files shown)


kalessil:
At the first look test cases are reported as it was intended. Which of pushed test cases you are referencing to? And for a bug-fix, I'd suggest using this PR as well.

Garethp:
In that case I wonder if there's something buggy with my set up. On my machine, the `empty()` inside `isNullableInt` throws a weak warning about how `empty` returns true on too many issues, as opposed to the suggestion of using `=== null`. Debugging in to it, it looked like it was because the resolved types of `$int` was empty, so it wasn't detecting that it was a nullable integer. Which is weird, because it detected the return types of functions fine.

If it's passing for you, I'll look at also installing the recommended version of PHPStorm (I'm on the latest) and using the PHP lib from that, and filing a bug report to Jetbrains if that solves it for me.

Garethp:
To clarify, I've written the test cases to report as they are supposed to, but because of that the tests don't pass on my development environment, due to the mentioned bug in `isIntNull` for me

kalessil:
I'll try running tests from this feature branch today evening, will be back then.

kalessil:
@Garethp : same for me:

```log
empty-function.php: Extra (14:12/11) :'empty($int)' ('empty(...)' counts too many values as empty, consider refactoring with type sensitive checks.) [HighlightInfoTypeImpl[severity=WEAK WARNING, key=INFO_ATTRIBUTES]]
empty-function.php: Extra (18:6/26) :'empty(getNullableString())' ('empty(...)' counts too many values as empty, consider refactoring with type sensitive checks.) [HighlightInfoTypeImpl[severity=WEAK WARNING, key=INFO_ATTRIBUTES]]
empty-function.php: Missing (14:12/11) :'empty($int)' (You should probably use 'null === $...'.)
empty-function.php: Missing (18:6/26) :'empty(getNullableString())' (You should probably use 'null === $...'.)
```

Garethp:
I get the first one, but not the second. Could you please send your php plugin version? I'll download it and see if it's a difference between the two, then I'll report this bug to Jetbrains. I think though that the test is working as expected, it's just a bug with the jetbrains PHP Type Resolver.

kalessil:
Thank you @Garethp, merging. Once tests are green again I will be on the travis PR.

PR #218, opened 2017-03-30
Multiple return statements

kalessil:


kalessil:
@artspb, @funivan : performance is very good, thanks for hints and implementation!

Issue #188, opened 2017-03-16
[unused variables] Report unused use() variable


rentalhost:
**See next example:**

```php
$a = null;
$b = function () use ($a) {
};
```

Note that `$b` uses `$a` only "virtually", but it is never readed inside it. I guess that in this case it should be represented like a `unused referenced variable`.

kalessil:
That's sweet, a good case =)

kalessil:
Added, reports unsed use-variables and reports only-write cases for the variables

rentalhost:
@kalessil No binary? 

Did you checked if it will warn too:

```php
$b = function () use (&$a) {
};
```

kalessil:
Fresh binary pushed now. And pretty similar cases is tested also.

rentalhost:
@kalessil I tried last binary, but unfortunetelly the `&$a` doesn't is reported. 😢

rentalhost:
Off-topic: there some doc that I can read to try to help with some easy-picks PRs? I am not a Java programmer, but I think that I can help in some cases, but I don't know from where I should starts.

kalessil:
Right, added the test case; fix will be available today.

Yep, docs are available, you can find  it [here](https://github.com/kalessil/phpinspectionsea/blob/master/docs/DEVELOPERS.md).

kalessil:
I'ts pretty easy to add tests, we have lists of missing tests [here](https://github.com/kalessil/phpinspectionsea/blob/master/RULES.md). Some text samples even [available](https://github.com/kalessil/phpinspectionsea/tree/master/php).

rentalhost:
@kalessil  Can you share your Java code style settings?

kalessil:
I use defaults, except aligning sequential assignments manually.

rentalhost:
Auto alignments doesn't works?

kalessil:
I didn;t find where I can switch this on...

rentalhost:
Seems that the align could be done by enable the "group declarations" on Java code style / Wrapping and Braces.

---

Okay, I have done the initial setup and I can run the tests, but I received 13 failed tests. I think that we can have some problem here, because in some comment you told that all is green. Or maybe I'm just forgetting something.

**For instance:**

> com.intellij.rt.execution.junit.FileComparisonFailure: usage-of-silence-operator.php: Missing (4:5/1) :'@' (Try to avoid using the @, as it hides problems and complicates troubleshooting.)

**Differences:**

[...]

PR #237, opened 2017-04-06
[#188] OnlyWritesOnParameterInspector: check if referenced variables are use()d too

rentalhost:
What I did was basically removes the `if()` that avoids any validation over _referenced variables_, before it checks if it was used inside the function. 

Now the `analyzeAndReturnUsagesCount` will receives an observation that the passed variable is a _referenced variable_. It'll allows that this method checks if the variable is used internally, and independently of it be a reference, it'll returns 0 if it is never used (registering the problem). After that, if the variable is a referenced variable, then all method could be avoided (once that is useful only for local variables).

Added some tests to make sure that I don't broke anything.

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/OnlyWritesOnParameterInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/OnlyWritesOnParameterInspector.java
@@ -40,7 +40,8 @@ public class OnlyWritesOnParameterInspector extends BasePhpInspection {
         return "OnlyWritesOnParameterInspection";
     }
 
-    static private PhpAccessVariableInstruction[] getVariablesAccessInstructions(String parameterName, PhpScopeHolder objScopeHolder) {
+    @NotNull
+    static private PhpAccessVariableInstruction[] getVariablesAccessInstructions(@NotNull String parameterName, @NotNull PhpScopeHolder objScopeHolder) {
         PhpEntryPointInstruction objEntryPoint = objScopeHolder.getControlFlow().getEntryPoint();
         return PhpControlFlowUtil.getFollowingVariableAccessInstructions(objEntryPoint, parameterName, false);
     }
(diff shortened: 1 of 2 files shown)


src/test/resources/fixtures/deadCode/parameters-writes-only.php
```diff
@@ -2,23 +2,31 @@
...
     public function method1(array $in, array &$out, $log) {
-        return
+        $outc = 1;
+        $function =
             function
                 (array $inner, array &$outer)
                 use (
                     <weak_warning descr="The variable seems to be not used.">$in</weak_warning>,
-                    &$out,
+                    &<weak_warning descr="The variable seems to be not used.">$out</weak_warning>,
```
kalessil: Too many changes here, it should be enough `&<weak_warning descr="The variable seems to be not used.">$out</weak_warning>,` only

rentalhost: I have created more tests on this file to make sure that other cases are not generating false-positives (it help me on manual tests step).

`&$out` was declared on `use()` but not on parent context, and never used (then _warn_);
`&$outb` was declared on `use()`, but not on parent context, but was used (then _ok_);
`&$outc` was declared on `use()`, and too on parent context, and too used (then _ok_);

kalessil: I see and It;s totally ok for manual, just the final tests cases needs to be minimalistic.

kalessil:
Well, extension of methods with boolean flags is not a good idea: the method already does 2 things (report and counting usages). And the boolean flag makes possible to invalidate method purpose at all, it will be difficult to maintain such solution - I did this as well in past, cost me much maintanance.

See how I would fix this (on original master: VCS -> Apply patch)
[__188__OnlyWritesOnParameterInspector__more_tests__bug_fixing.patch.txt](https://github.com/kalessil/phpinspectionsea/files/904534/__188__OnlyWritesOnParameterInspector__more_tests__bug_fixing.patch.txt)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/OnlyWritesOnParameterInspector.java
```diff
@@ -40,6 +40,11 @@ public String getShortName() {
         return "OnlyWritesOnParameterInspection";
     }
 
+    static private PhpAccessVariableInstruction[] getVariablesAccessInstructions(String parameterName, PhpScopeHolder objScopeHolder) {
```
kalessil: Good idea to make it static =)

rentalhost:
Okay, I did the needly changes. 

Basically, I took the responsability of count variable usage to the static method `getVariablesAccessInstructions`, which the `analyzeAndReturnUsagesCount` method will use now too. 

It is need because the inspection over referenced variable too need of this data. The advantage is that this inspection will not need the full implementation of `analyzeAndReturnUsagesCount` (that keep being useful for common variables).

kalessil:
Thank you @rentalhost

PR #241, opened 2017-04-09
Ignore/Remove files from git index and upate docs

davidmpaz:
Remove files that are updated after project configuration
setup in ".idea" directory. Update dev docs and add seed
for log4j settings.

Changes:

* Ignore/Remove files that are updated after project configuration
  setup inside .idea directory.
* Add file with basic default settings for log4j to use in Test setup.
* Add dictionary file to docs.
* Rename module file to PhpInspectionsEA.iml.dist. Update configuration docs.
* Update dev docs:
    - Move up, step "setting up libraries". Setting up "Module" step make more sense after setting up "Libraries"
    - Add more dev docs on setting up project.

----
The reason why I am requesting this PR is that, as newcomer, when finished all setup, i had several configuration files modified inside `.idea` dir. I think git repository should remain clean and not affected by these changes, we should favor adding docs on how to set up instead on sharing local development environments. 

Those files that get changed every time a new developer adapt project to his/her environment should be ignored and added as sample file or alike to be modified explicitly. And documentation state what need to be done to configure.

I have tried to add documentation on setting up all features I saw that got updated for me. If i missed something i can add it if you give some hint of missing setup steps.

Cheers,
David


kalessil:
LGTM, thank you @davidmpaz

kalessil:
@funivan @rentalhost : be sure to backup your workspace and idea files before pulling master changes - this PR does cleanup and updates docs.

funivan:
@kalessil ok =)

Issue #239, opened 2017-04-07
Suggests reverts simple conditions with else-only


rentalhost:
Hard to explain, but consider the next code:

```php
if (!expr()) {
    doA();
}
else {
    doB();
}
```

It could be writed as:

```php
if (expr()) {   // Removed the not-operator
    doB();      // The previous "else" statements
}
else {
    doA();      // The previous "if" statements
}
```

It is... keeps the `if()` always positive. It'll help a lot with simple ifs conditionals. I guess that I see that in some IDE (maybe for C#).

**Edit:** ah, yeah. The own PS allows you flip if-else statements. Then I guess that need justs an inspection. If possible, uses the own QF from PS.

funivan:
1. Flip if else is standard phpstorm intention
2. For me it would be interesting to find ` NOT NOT ` conditions. 
for example:
```php
!$dealer->isInactive();
!$dealer->withoutEmail();
!isInvalid();
```
And suggest user to rewrite this conditions, because is is really hard to understand them.

rentalhost:
I guess that your examples will works fine if applied in a general way too. Suggests flip for both cases will make easier to understand, `!$dealer->isActive()` or `!$dealer->isInactive()`. It will avoid a mapping for negative words.

rentalhost:
Closing: PR #249

PR #249, opened 2017-04-14
[#239] Suggest to avoid not-operator on conditionals

rentalhost:
I have implemented a suggestion to avoid not-operator in simple conditionals, as described on #239.

- [x] Inspection implemented with UTs;
- [x] Quickfix implemented with QFTs;

What it basically does is warn dev about use not-operator in cases where he just can swap position of if-else statements, avoiding this operator. It make code easy to understand and not change behavior.


rentalhost:
**Post note:**

The unique problem that I found is that the `PsiElement.replace()` forces the formatting of the `PsiElement` that will be the replacement. But the same happen when you trigger "flip if-else" from PS. For instance, if I have:

```php
if (!true) { leftSide(); } else { rightSide(); }
```

I expects that after quickfix it be:

```php
if (true) { rightSide(); } else { leftSide(); }
```

But this method forces it to be:

```php
if (true) {
    rightSide();
} else {
    leftSide();
}
```

Is is bad? No. 
But I'm quite upset that a simple replacement forces formatting.

src/test/resources/fixtures/ifs/if-avoid-not-operator.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+if (<weak_warning descr="This negative if conditional could be avoided">!expr()</weak_warning>) { a(); b(); }
+else { c(); d(); }
+
+if (!true) { a(); b(); }
```
funivan: I thinks this statement is complicated and should be ignored.

kalessil:
I need to finish current release before I can do any reviews.

src/test/resources/fixtures/ifs/if-avoid-not-operator.php
rentalhost: This is partially ignored. On  reality, the if() itself is ignored, then the last elseif() branch is inspectioned. Futurely I hope could treats complex cases like that, but for now it will works fine. Have you some code in mind where it will be a problem?

funivan: For last `elseif` all is ok =) It will be cool to test this inspection in real code base.

rentalhost: I have tried on vendor folder on a Laravel project and I have found a lot of compatible cases,  none false positive. :)

kalessil:
This week will have time to review this.

rentalhost:
Code rebased to head, some small code improvements (code style, basically).

kalessil:
Super, thank you so much for working out PRs queue!

kalessil:
Thank you @rentalhost !

Issue #260, opened 2017-04-18
Activities coordination


kalessil:
Here I want to structure my vision of activities around the plugin:

- Q2 2017: stabilization, crowdfunding obligations
  - bug-fixes
  - QFes
  - dev-stuff (graddle, Jenkins)
- Q3 2017: further stabilization
  - key features hardening (probable bugs, NPEs, not optimal if conditions)
  - next battle for reasonable performance
  - backlog big cleanup
- Q4 2017: to be planned

Small features will be picked up, but nothing fundamental - I expect a reduction of OSS availability starting Q2.

kalessil:
@rentalhost @funivan @davidmpaz @Garethp  : I want to define some focus in my activities and understand how to align our activities.

Would you please share your thoughts - are there any ways for you to work on some topics together or perhaps I have to dedicate fixed time for some activities?

rentalhost:
If you mean about issues, I like to work on issues that I have reported and QFs/UQFs for current features (based on RULES.md). Maybe first QFs to respect Q2. And I suggest that, if possible, each one could do the same, then we can reduce the issues list a lot (current 61 items) and create a lot of QFs.

funivan:
@kalessil totally agree. I think we should reduce number of new inspections and try to close all `false-positive`

What i`m planing to do. I have a lot of interesting ideas for new inspections. But i will try to implement them in the separate plugin and test it in my project. Q3 or Q4 can be a good period to push some inspections to this plugin (step by step)
In this repository I will try to fix some bugs and write some  QF  (I love them))

rentalhost:
Ah, a good idea that I see in some projects, like Laravel, is separates the issue to a `internals` project (something like _kalessil/internals_ or _kalessil/phpinspectionsea-internals_). Then we could discuss what we could do in each case before create an issue on this project. On this project, only issues "closest to WIP" will remain. For instance, in last day I sent about 5 issues, but I guess that we could discuss before turn it a "priority", but currently we have no "space" for that. It'll reduce a lot the issues and keep only "currently important issues". The major problem is that the people that don't know how it works probability will create an issue directly on this project, and I don't know if we can move to internals on github.

kalessil:
It was lying on the surface, how I could miss that idea ). 
Community edition (CE) version of the plugin should resolve current bottleneck issues (back-merge is another topic, but if each of new inspections is self-contained we are on the safe side).

Should I fork a new project (propose name please) and create a new plugin for it?

[...]

PR #267, opened 2017-04-20
Reviewing the RULES.md;

rentalhost:
Following #260, I proposes to review fully the `RULES.md` file (_currently it is a separated file called `RULES.dev.md` to not override current file_), then we could identify where we can improve that. 

**I too did some changes on this file:**

* Removed the **Group** column, because the own table title specify that;
* Replaced the **Short Name** column to **Class**, then instead of write the short name ("`*Inspection`"), we specify directly the class name that treats the inspection ("`*Inspector`");
* Renamed the **Full Name** column to **Short description**;
* Removed the **QFTs** column, now the own **QF** column will treat that too (_see next_);
* Align the columns **UTs**, **QFs** and **Doc** on center;

**Then I have implemented some good pratices on file:**

* The cell value `yes` mean that the item is fully implemented;
* The cell value `yes*` mean that the item is fully implemented, but not the unit tests (_useful for QFs_);
* The cell value `no` mean that the item was not implemented, but is possible;
* The cell value `-` mean that the item can't be implemented (_impossible_);

---

Well, the table cells is empty because I suggests too review "_over the top_" each inspection to take care with:

* Small fixes, for instance, add `final` where possible;
* Code style/realignments (_I will works on a good code style soon_);
* Separate multiple inspections wherever possible;
* Improve and create comments in obscure codes;
* Implements more tests and code coverage;
* Implements the quick-fixes if possible and you can, with tests;
* Implements documentations;
* File copyright;
* Small optimizations, so on;

### DIFF
--- a/RULES.dev.md
+++ b/RULES.dev.md
@@ -1,0 +1,230 @@
+# Inspections list
+
+> **Attention:** this inspection list is in development.
+> For stable list, check [RULES.md](https://github.com/kalessil/phpinspectionsea/blob/master/RULES.md) file. If you wants help with this list, read the [PR #267](https://github.com/kalessil/phpinspectionsea/pull/267).
+
+**Abbreviations meaning:**
+
+* **UTs**: unit tests for the inspection;
+* **QF**: quick-fix (`*` *mean not-tested yet*);
+* **Doc**: fully documented at [documentation page](docs/inspections.md);
+
+**Values meaning:**
+
+* `yes`: fully implemented;
+* `yes*`: fully implemented but not tested programatically (*unit tests*);
+* `no`: yet not implemented, but possible;
+* `-`: impossible to implements;
+
+---
+
+## Type Compatibility
+
+| Class                                       | Short description                                   | UTs | QFs  | Doc |
+| :------------------------------------------ | :-------------------------------------------------- | :-: | :--: | :-: |
+| IsEmptyFunctionUsageInspector               | 'empty(...)' usage                                  |     |      |     |
+| TypeUnsafeComparisonInspector               | Type unsafe comparison                              |     |      |     |
+| TypeUnsafeArraySearchInspector              | 'in_array(...)', 'array_search()' type unsafe usage |     |      |     |
+| ForeachSourceInspector                      | Foreach source to iterate over                      |     |      |     |
+| GenericObjectTypeUsageInspector             | Usage of object type in PhpDoc                      |     |      |     |
+| ArrayTypeOfParameterByDefaultValueInspector | Parameter could be declared as array                |     |      |     |
+
+## Control Flow
+
+| Class                                    | Short description                                        | UTs | QFs  | Doc |
+| :--------------------------------------- | :------------------------------------------------------- | :-: | :--: | :-: |
+| UnSafeIsSetOverArrayInspector            | 'isset(...)' usage                                       |     |      |     |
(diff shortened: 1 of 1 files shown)


funivan:
1. Should we keep non-implemented inspections in this list?
2. Why we should place "inspector" instead of "inspection"? Inspection name is stored in XML file (phpstorm settings)
3. I suggest to autogenerate this file. ;) Take a look at https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/README.rst you can find rules list at the bottom.

rentalhost:
1. No, because it should be used as reference on future for yet implemented inspections;
2. On reality, I don't know why we have two different names. I guess that it must be normalized. Both class and XML turn to `Inspector` or `Inspection`. Except if it some kind of limitation;
3. Should be interesting, but I don't know how it works;

funivan:
2. We can change class name without any side effects for the end users.

funivan:
+ i do not agree with `yes*` hard to understand for end users ;)

rentalhost:
The focus is the development, because the goal is implements any possibility of QF. Then is like a "temporary status" to we can understand "what is pending";

funivan:
@rentalhost I think that `RULES.md` is for end users. Probably we cant create QF for all inspections. So `yes` can be placed forever for some inpspections.

rentalhost:
This is the reason to we work, for now, on `RULES.dev.md`, then on end we can just rename that to `RULES.md`. The `yes*` is just a control to make sure that "_quick-fix exists, but have not unit tests_". As some inspections will never have a QF because is impossible (for instance, the "_foreach iterable_"), then we just mark as `-` (currently it is marked as `n/a`).

kalessil:
RULES.md is for end-users (incl. us) and dropping the first column is a good idea.

Dropping QFTs column is not so good idea, makes my work harder =) 
How do I work (I'm  open for improvement proposals):
- Check if QFTs says no;
- Quick lookup for QF exists;
- Quick lookup for UTs exists;

And this allows me to go in iterations:
- Add QFs tests for existing once, fixing discovered bugs;
- Add QFs for tested inspections, improving user experience;
- Annoying part: no/partial tests => bug-fixing, QF implementation, QFTs implementation and bug-fixing;

QFTs column is an entry point, allows to reduce cognitive load for my tired brain =)

kalessil:
And we need an auto-generator (can be PHP-based =)):
- which reads the existing file and remembers UTs/QF/QFTs/Doc values
- generates a new file from [here](https://github.com/kalessil/phpinspectionsea/blob/master/META-INF/plugin.xml)

kalessil:
@rentalhost : can be auto-generator part of this PR?

rentalhost:
Sure. But I don't know how it works. The PHP part is ok, but I don't know about Java, because probably it should read Java code to generate.

davidmpaz:
If we can extract all the relevant information from plugin.xml. We could even use an xslt transformation to obtain the kind of table (html?) we need.

funivan:
plugin.xml doesn't know anything about documentation, quick fix an tests.

kalessil:
I would imagine following logic:

- parse existing RULES.md, extract details about the individual inspections (UTs, QF, QFTs, Docs)
- read the xml (inspection, group, title)
- generate new RULES.md from the xml, injecting details from the first step

rentalhost:
I'm closing it for now.
Lets think about make it automatic, but discuss first on an issue.

Issue #277, opened 2017-04-25
Optimize JPanel's


rentalhost:
Following #276 idea, I like to do the same for JPanel created by some inspections. Currently each option is manually declared on the own inspection class. I don't know if we can change that, but we can make the readability better.

And it is my test result:

```java
public JComponent createOptionsPanel() {
    return OptionsComponent.create((component) -> {
        component.createCheckbox("Report empty() usage", 
            this.optionEmptyUsage, 
            value -> this.optionEmptyUsage = value);

        component.createCheckbox("Suggest to use count()-comparison", 
            this.optionSuggestUseCount, 
            value -> this.optionSuggestUseCount = value);

        component.createCheckbox("Suggest to use null-comparison", 
            this.optionSuggestToUseNullComparison, 
            value -> this.optionSuggestToUseNullComparison = value);
    });
}
```

As I am not a Java programmer, then I have some limitations. For instance, the lambda parameter `component` receives an instance of  `OptionComponent`, that will be used to create each option on inspection. But it receives three parameters: the option title, the default value and a lambda that will be called when the option is changed on IDE. I don't know if I can optimize that to receive only a reference of default value (seems that Java don't have something like pass-as-reference).

funivan:
Sometimes panels are complicated. And it is better to create external class.

rentalhost:
I am working on that just now. I have finished, now I am having some issues as, for instance, the generated checkbox works but not affects or save preferences. haha I am investigating it then I will make a PR about ~1h (else, I will continues tomorrow).

rentalhost:
Argh... I can't do it work. As Java have no support to reference, then I need create an Atomic type (or Mutable), but too for some reason, it change the instance value correctly but IDE can't see that. Each time that I open the settings it creates a new instance (even when it is static!). I'll push to someone check the current code, but for now I will not make a PR.

rentalhost:
About the issue that I have mentined before, I found the cause. Seems that the Intellij SDK can handle only configurations of type `boolean` and `List<>`, and I am trying to use `AtomicBoolean` and similar (if it was possible, this would greatly reduce the code as well).

**From (what I have did after found this problem):**

```java
// boolean optionAnalyzeInterfaces
component.createCheckbox("Analyze interfaces", 
    optionAnalyzeInterfaces, 
    (isSelected) -> optionAnalyzeInterfaces = isSelected);
```

**To (what I liked to do, but don't worked):**

```java
[...]

PR #283, opened 2017-04-26
[#277] Optimizes the JComponent's creation on all inspection classes

rentalhost:
_Okay, it modifies 33 files by a good reason: **readability**.
Future devs will appreciate it a lot._

Following #277, I did this changes to optimize the `JComponent`'s creation on inspections that declares it. It is inevitably part of the inspection class, generated by method `createOptionsPanel()`.

Basically I have create a new class called `OptionsComponent` that can handle tree types of components: _checkboxes_ (via `createCheckbox()`), _radios buttons_ (via `createRadio()`) and _lists_ (via `createList()`).

~~As exception (_included on original code_), the _radio buttons_ is treated like _checkboxes_. It happen only on inspection "_yoda/regular conditions style usage_", and in this case, we need have the possibility to disable this inspection by unchecking both options (_yoda_ and _regular style_). We can create real radio buttons here instead (via `JRadioButton`), but I don't think need and checkbox works for now.~~ Fixed/waiting for #288.

~~Another change that I have did was rename all fields that is treated like an option. Then instead it seems like an static variable (like `SUGGEST_TO_USE_COUNT_CHECK`) now it is treated like an option variable (like `optionSuggestToUseCountCheck`).~~ Reverted to avoid BC (will be discussed later).

In general, it have reduced the code related to inspection options from about 260 lines to 45 (excluding spaces). Which is a reduction of 83%, which means that I will pay less when do a new eye exam. It mean that a lot of duplicate/similar code and needless inner classes was dropped too.

_Well, the next step is do a similar thing with UTs._

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
@@ -34,8 +34,8 @@ import org.jetbrains.annotations.NotNull;
 
 public class ComparisonOperandsOrderInspector extends BasePhpInspection {
     // Inspection options.
-    public boolean optionPreferYodaStyle    = false;
-    public boolean optionPreferRegularStyle = false;
+    public boolean PREFER_YODA_STYLE    = false;
+    public boolean PREFER_REGULAR_STYLE = false;
 
     final static private String messageUseYoda    = "Yoda conditions style should be used instead";
     final static private String messageUseRegular = "Regular conditions style should be used instead";
@@ -58,7 +58,7 @@ public class ComparisonOperandsOrderInspector extends BasePhpInspection {
     public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, boolean isOnTheFly) {
         return new BasePhpElementVisitor() {
             public void visitPhpBinaryExpression(BinaryExpression expression) {
-                if (!optionPreferYodaStyle && !optionPreferRegularStyle) {
+                if (!PREFER_YODA_STYLE && !PREFER_REGULAR_STYLE) {
                     return;
                 }
 
@@ -83,11 +83,11 @@ public class ComparisonOperandsOrderInspector extends BasePhpInspection {
                     return;
                 }
 
-                if (optionPreferYodaStyle && isRightConstant) {
+                if (PREFER_YODA_STYLE && isRightConstant) {
                     holder.registerProblem(expression, messageUseYoda, ProblemHighlightType.WEAK_WARNING, new TheLocalFix());
                     return;
                 }
-                if (optionPreferRegularStyle && isLeftConstant) {
+                if (PREFER_REGULAR_STYLE && isLeftConstant) {
                     holder.registerProblem(expression, messageUseRegular, ProblemHighlightType.WEAK_WARNING, new TheLocalFix());
                 }
             }
@@ -97,8 +97,8 @@ public class ComparisonOperandsOrderInspector extends BasePhpInspection {
     public JComponent createOptionsPanel() {
         return OptionsComponent.create((component) -> {
             component.createRadio((radioComponent) -> {
-                radioComponent.createOption("Prefer yoda style", optionPreferYodaStyle, (isSelected) -> optionPreferYodaStyle = isSelected);
-                radioComponent.createOption("Prefer regular style", optionPreferRegularStyle, (isSelected) -> optionPreferRegularStyle = isSelected);
+                radioComponent.createOption("Prefer yoda style", PREFER_YODA_STYLE, (isSelected) -> PREFER_YODA_STYLE = isSelected);
+                radioComponent.createOption("Prefer regular style", PREFER_REGULAR_STYLE, (isSelected) -> PREFER_REGULAR_STYLE = isSelected);
             });
         });
     }
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
@@ -15,12 +15,13 @@ import com.jetbrains.php.lang.psi.elements.ConstantReference;
 import com.jetbrains.php.lang.psi.elements.StringLiteralExpression;
 import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpElementVisitor;
 import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpInspection;
-import net.miginfocom.swing.MigLayout;
-import org.jetbrains.annotations.NotNull;
+import com.kalessil.phpStorm.phpInspectionsEA.options.OptionsComponent;
 
 import javax.swing.*;
 import java.util.HashSet;
 import java.util.Set;
+
+import org.jetbrains.annotations.NotNull;
 
 /*
  * This file is part of the Php Inspections (EA Extended) package.
@@ -32,10 +33,9 @@ import java.util.Set;
  */
 
 public class ComparisonOperandsOrderInspector extends BasePhpInspection {
-    // configuration flags automatically saved by IDE
-    public boolean CONFIGURED           = false;
-    public boolean PREFER_YODA_STYLE    = false;
-    public boolean PREFER_REGULAR_STYLE = false;
+    // Inspection options.
+    public boolean optionPreferYodaStyle    = false;
+    public boolean optionPreferRegularStyle = false;
 
     final static private String messageUseYoda    = "Yoda conditions style should be used instead";
     final static private String messageUseRegular = "Regular conditions style should be used instead";
@@ -58,10 +58,14 @@ public class ComparisonOperandsOrderInspector extends BasePhpInspection {
     public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, boolean isOnTheFly) {
         return new BasePhpElementVisitor() {
             public void visitPhpBinaryExpression(BinaryExpression expression) {
+                if (!optionPreferYodaStyle && !optionPreferRegularStyle) {
+                    return;
+                }
+
                 /* verify general structure */
-                final IElementType operator = CONFIGURED ? expression.getOperationType() : null;
-                final PsiElement left       = CONFIGURED ? expression.getLeftOperand() : null;
-                final PsiElement right      = CONFIGURED ? expression.getRightOperand() : null;
+                final IElementType operator = expression.getOperationType();
+                final PsiElement   left     = expression.getLeftOperand();
+                final PsiElement   right    = expression.getRightOperand();
                 if (null == operator || null == left || null == right || !operations.contains(operator)) {
                     return;
                 }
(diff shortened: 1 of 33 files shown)


src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/UsageOfSilenceOperatorInspector.java
```diff
@@ -33,8 +34,8 @@
  */
 
 public class UsageOfSilenceOperatorInspector extends BasePhpInspection {
-    // configuration flags automatically saved by IDE
-    public boolean RESPECT_CONTEXT = true;
+    // Inspection options.
+    public boolean optionRespectContext = false;
```
kalessil: Will destroy custom configurations, names must not change. Know naming is weird, but i meant this naming to spot magically saved options, sorry.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
```diff
@@ -79,60 +83,24 @@ public void visitPhpBinaryExpression(BinaryExpression expression) {
...
-            preferRegularStyle.addChangeListener(e -> {
-                PREFER_REGULAR_STYLE = preferRegularStyle.isSelected();
-                if (PREFER_REGULAR_STYLE) {
-                    preferYodaStyle.setSelected(false);
-                    PREFER_YODA_STYLE = false;
-                }
-
-                CONFIGURED = PREFER_YODA_STYLE || PREFER_REGULAR_STYLE;
+        return OptionsComponent.create((component) -> {
+            component.createRadio((radioComponent) -> {
```
kalessil: Elegant! We need 3rd one: "No preferences" -> CONFIGURED

kalessil:
Elegant, love it.

We need only to recover fields names, as refactoring destroys users settings - nogo. +See 2 other comments.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/UsageOfSilenceOperatorInspector.java
rentalhost: Yeah, I have checked that. But there are just few customizable inspections (_about 16 inspections_), then I belive that the user affected will be a small part. And from here to future we can keep the naming convension well.

About _spot magically saved options_, it is just the reason that I prefixed `option`. Then we know that `if(optionX)` is mean to be customizable "X" option. Use a uppercased variable seems that we working with a constant, and make confuses on first step.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
rentalhost: On original code we have only two options. The "configured" was a mix of both. Something like `configured = a.isSelected() || b.isSelected()`. I really think about create a 3rd one to treats a "disabled option", and make more sense too (then we can use a `JRadioButton` instead too to make yet more sense). 

By other side, not make sense user enable the inspection and disable it internally. I guess that we should consider mark some style by default, then user can decide to disable it "as inspection" and not by radio.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/UsageOfSilenceOperatorInspector.java
kalessil: @rentalhost : this is really needed. I need a good analogy, imagine that yo was tweaking IDE code style (e.g. align assignments, + 25 other options you are not remembering anymore and) and PS on update did reset them.

I don't want our community to run into this. The change is a good one, but we can not afford it, pleaseeeeeeeeeeee return naming back.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
kalessil: Good idea lets keep 2 of them and deactivate the inspection by default.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
```diff
@@ -0,0 +1,163 @@
...
+    public void createRadio(@NotNull final Consumer<RadioComponent> componentDefiner) {
+        final RadioComponent radioComponent = new RadioComponent();
+        componentDefiner.accept(radioComponent);
+    }
+
+    /**
+     * Creates a List.
+     *
+     * @param fieldTitle  The title of this field on panel.
+     * @param listItems   The default or current list items.
```
funivan: How it can be `default or current` - this listItems should be always a list of items or i`m wrong?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
```diff
@@ -0,0 +1,163 @@
...
+
+            if (defaultValue) {
+                currentSelection = createElement;
+            }
+        }
+
+        /**
+         * Handle a JCheckBox component and it Update Consumer.
+         */
+        private class Element {
```
funivan: Maybe we can rename class to CheckboxElement?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
```diff
@@ -0,0 +1,163 @@
...
+             * @param updateConsumer The update consumer (called after item change).
+             */
+            Element(final String fieldTitle, final boolean defaultValue, final Consumer<Boolean> updateConsumer) {
+                this.updateConsumer = updateConsumer;
+
+                checkbox = new JCheckBox(fieldTitle, defaultValue);
+                checkbox.addItemListener((itemEvent) -> {
+                    final boolean isSelected = checkbox.isSelected();
+
+                    if (currentSelection != null && currentSelection != this) {
```
funivan: `null!=currentSelection ` ;)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/UsageOfSilenceOperatorInspector.java
rentalhost: Okay, then I will undo that. But, can we reconsider that for 3.x, as a breaking-change? Maybe next year.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
rentalhost: Once opted for this, we should consider that it will have _default checked option_ when this inspection is enabled. I don't think that is need deactivate this inspection by default. Then... could we consider the use the regular style here? The yoda style is more relevant on C language. (Except if you really prefer disable it, but for a big surprise, I don't know about this feature until I work on it, and I was really willing to open an issue to implement it)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
rentalhost: Initially, I tried to use the `AtomicBoolean` that initially receives the "default value" defined on field of inspection class. Then for all components I create this parameter ("default or current"). Next, I found that it is impossible currently, because IDE don't recognize the `AtomicBoolean` as a configurable option (_it just ignore it_).

The unique type that we can have as a _non-primitive type_ is a `List<>`. Then it could have the default value (which is defined from inspection class) or the current value (after IDE load that). It are "magic fields" which are strangely managed by IDE (rather than doing it more efficiently).

To summarize: this variable is used both to read the default values and to read the current configuration.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
rentalhost: I don't know what is the best name for this case. Note that it is not really a checkbox, not even a radio button. It stores a checkbox and this consumer. Then the `setSelected()` method can both modify the checkbox checked status and trigger the consumer.

But, I do agree that `Element` is not a good name.
New suggestions?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
rentalhost: Weeeeelll.... I don't agree. In all project code we have a mix of two styles: Yoda and regular.
But let talk about that on this topic: #286.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
kalessil: Widget?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
kalessil: Remembering related discussions, disabling it by default is the most non-conflicting approach. PHP community is pretty aggressive at this specific check (have no idea).

And please make yoda style the default one.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/UsageOfSilenceOperatorInspector.java
kalessil: To do it nicely we need to have transition version with new and old properties and copy over values.
And at some point (6 months later) we can drop old variables.

If you transition version can be done in the PR, we can start with the next release already.

kalessil: Let do panel migration first with the new and nice solution. 2nd PR can be migration.

rentalhost: Yes, I do agree that is better implements it first (with original variable names), then care about variable changes in another context (it can be done in another moment, another PR). It because I need see if it is possible currently too (for now I can think about a workaround only, and I like to avoid that).

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
rentalhost: Okay, then it should be disabled by default (the inspection itself), but when enabled we should mark Yoda as default style. Right?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/options/OptionsComponent.java
rentalhost: I guess that we can work with `Widget`. Noted!

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ComparisonOperandsOrderInspector.java
rentalhost: In all way, I should open a new issue because it is not related to this PR. #288

rentalhost:
Done all things and rebased to master.

Tests passed successfully.
Tested manually on PS.
No breaking-changes.

Best option to merge (if have no more issues) is squash.

kalessil:
Looks good, will merge today.

rentalhost:
@kalessil Some news here?
I can rebase it to head again before merge it.

kalessil:
Working thru PRs last week, so will be merged soon.

Rebase is a good idea - builds will be triggered then.

rentalhost:
Rebasing done without issues.

kalessil:
Super, I'll take care about merging.

kalessil:
Thank you @rentalhost.

PR #308, opened 2017-05-15
[#73] Warn if default value is passed for last arguments

### LINKED ISSUE #73: [new] Warn if default value is passed for last arguments
This should checks if you are passing the default value of a method. But it should occur only if it's explicit. For instance:

```
function method($a = 0, $b = 0) {
}
```

Then:

```
method();     // Valid.
method(1);    // Valid.
method(0);    // Warn: you can omit the LAST argument (+fix).
method(1, 0); // Warn: you can omit the LAST argument (+fix).
method(0, 0); // Warn: you can omit the LASTS TWO arguments (+fix).
method(0, 1); // Valid.
```

Unfortunatelly PHP have no `default` keyword, then the last should be accept too. 😄

rentalhost:
Fixes #73.

It basically will check if the lasts arguments on a call is equals to the default value, if true, will suggests to drop it (will not affects the expected behaviour).

If you have a method with two parameters, for instance, `method($a = null, $b = null)`, it should warning you if you call like `method(null, null)` for last two arguments. But it should accepts if your last argument doesn't matches with default value, for instance, `method(null, 123)` is valid (not trigger the inspection) in any parameter.

**Todo:**

* [x] Inspection;
* [x] UTs;
* [x] QFs/QFTs;
* [x] Text review (_someone, please?_) - mean, the warning and quickfix message, rules and description;


src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
```diff
@@ -0,0 +1,140 @@
...
+ * This file is part of the Php Inspections (EA Extended) package.
+ *
+ * (c) Vladimir Reznichenko <kalessil@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+public class ParameterEqualsDefaultValueInspector extends BasePhpInspection {
+    private static final String message = "This parameter could be dropped, because the value is the same from default value.";
```
funivan: I think we should change word `parameter` to the `argument`
There is big difference: http://stackoverflow.com/questions/156767/whats-the-difference-between-an-argument-and-a-parameter
Maybe this message will be better: `Argument can be safely dropped. Default value is the same `

META-INF/plugin.xml
```diff
@@ -620,6 +620,10 @@ We keep delivering funded results =)
         groupName="Code Style"                                    enabledByDefault="false" level="WEAK WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.ParameterDefaultValueIsNotNullInspector"/>
     <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
+        shortName="ParameterEqualsDefaultValueInspection"         displayName="Calls with the same default value"
```
kalessil: Would you please apply following changes: title = "Unnecessary arguments specification", group = "Unused".

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
kalessil: Good proposal, probably this will be clear enough: "The argument can be safely dropped, as identical to the default value."

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
```diff
@@ -0,0 +1,140 @@
...
+
+    @NotNull
+    public final String getShortName() {
+        return "ParameterEqualsDefaultValueInspection";
+    }
+
+    @NotNull
+    @Override
+    public final PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder problemsHolder, final boolean onTheFly) {
```
kalessil: Will resume to this today.

src/main/resources/inspectionDescriptions/ParameterEqualsDefaultValueInspection.html
```diff
@@ -0,0 +1,6 @@
+<html>
+    <body>
+        <p>Suggests to remove the parameter assignment if it is the same of declared as default.</p>
```
kalessil: Reports arguments which can be safely dropped, as they are identical to the default value.

src/test/resources/fixtures/codeStyle/parameters-default-equals.fixed.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+function method($a = 0, $b = 0) {
```
kalessil: - Method?
- Test also on abstract methods.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
```diff
@@ -0,0 +1,140 @@
...
+                    break;
+                }
+
+                if (referenceParameterLower != null) {
+                    final PsiElement referenceParameterLowerPrev = ((PhpPsiElement) referenceParameterLower).getPrevPsiSibling();
+                    final PsiElement referenceParameterStart = (referenceParameterLowerPrev == null)
+                                                               ? referenceParameterLower
+                                                               : referenceParameterLowerPrev.getNextSibling();
+
+                    problemsHolder.registerProblem(problemsHolder.getManager().createProblemDescriptor(
```
kalessil: Why so complicated?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
```diff
@@ -0,0 +1,140 @@
...
+                        referenceParameterLower,
+                        referenceParameters[referenceParameters.length - 1],
+                        message,
+                        ProblemHighlightType.WEAK_WARNING,
```
kalessil: I'd think of using `LIKE_UNUSED` instead - better visual clarification.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
```diff
@@ -0,0 +1,140 @@
...
+import com.jetbrains.php.lang.psi.elements.*;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpElementVisitor;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpInspection;
+
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * This file is part of the Php Inspections (EA Extended) package.
+ *
+ * (c) Vladimir Reznichenko <kalessil@gmail.com>
```
kalessil: Yep, don't be shy =)

rentalhost: Done! :)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
rentalhost: Done.

META-INF/plugin.xml
rentalhost: Done.

src/main/resources/inspectionDescriptions/ParameterEqualsDefaultValueInspection.html
rentalhost: Done.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
rentalhost: Done.

src/test/resources/fixtures/codeStyle/parameters-default-equals.fixed.php
rentalhost: * Replaced "method()" to "test()";
* Now testing method calls too (now I found 51 cases on my project!);

Done (pushing now).

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/ParameterEqualsDefaultValueInspector.java
rentalhost: My logic was: 

* L69: The looping is done from last argument to first.
* L76: It will "accept as unused" if the type and text from argument is the same from default value;
* L78: It will store the "first unused argument" (the last we just know, because is the last parameter);
* L79: then will keep searching for "even deeper equality".

---

* L86: If we found some equality;
* L92: Then we register the problem...
* L93: highlighting the first equals argument...
* L94: to last argument on call.

---

* L87: backing to this line, we too store the "comma", because the fix will drop it too.
* L88: if we have no previous argument, then it should drop from first to last; if we have previous argument, it will drop from comma immediately after the previous argument (that will not be dropped) to last argument;

kalessil:
The PR is gone, @rentalhost can you check why so please?

rentalhost:
@kalessil Ops! I deleted the branch unintentionally (I think I've mistaken another PR).

kalessil:
Thank you @rentalhost.

PS: I'll start using it in my env, depending on the personal feeling I might want to change reporting behaviour. So if you see some refactoring on your PRs, they are due to usability/future support reasons.
I often do change my inspections if don't feel comfortable during the daily usage.

rentalhost:
No problem, it is expected. :)

Issue #309, opened 2017-05-16
Callables in 'for' loop termination condition - Quick Fix?


Debeet:
Would it be possible to provide a Quick Fix for this scenario?

`for ($i = 0; $i < count($values); $i++)`
=>
`for ($i = 0, $iMax = count($values); $i < $iMax; $i++)`

rentalhost:
There are some problem to I work on it?
(In fact I did not see your assignee, and I'm already working on it.)

kalessil:
Nope, you can grab it =)

rentalhost:
Great! 

Can you explain in which case the [`messageExternal`](https://github.com/kalessil/phpinspectionsea/blob/fb95f2fe4fb713e3fa66233547d7aef1df48f529/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java#L42) is registered for? I mean, a real code to I test it. I am developing the UTs and QFs, and too the CC (this is the unique line that I can test for because I just not understand what it should be used).

kalessil:
It's used when inspection fires on the incomplete statement (during typing in), this probably can be dropped as inspections are fired only on finished statements.

But it worth double-checking against my code if it's the same case.

rentalhost:
I have done a lot of checks, and I don't found a case where it could be possible, then I dropped it for now. It seems be some [old code](https://github.com/ovr/phpinspectionsea-mirror/blob/master/src/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java).

rentalhost:
Closing: fixed on #313.

Debeet:
Thank you! 👍

PR #313, opened 2017-05-17
[#309] QF to callables at conditionals on for()

rentalhost:
Following #309 

* [x] Improved the code-style (eg. alignment, _space for the code to breathe_, etc.);
* [x] Dropped the `messageExternal` (because it is from an old code, [probably](https://github.com/kalessil/phpinspectionsea/issues/309#issuecomment-302102714));
* [x] Added UTs with full coverage;
* [x] Implemented the QF, QFTs and full coverage too;
* [x] Replaced the message to a more generic one, once that the QF will do the work from now;
* [x] Unknow variables are renamed to `$loopsMax` instead of `$loopsCount` (to be compatible with the "existent definition", as in `$iMax`);
* [ ] Text review is welcome (_but I not changed too much, I guess_);

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java
@@ -1,12 +1,20 @@
 package com.kalessil.phpStorm.phpInspectionsEA.inspectors.semanticalAnalysis;
 
+import com.intellij.codeInspection.LocalQuickFix;
+import com.intellij.codeInspection.ProblemDescriptor;
 import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.codeInspection.ProblemsHolder;
-import com.intellij.psi.PsiElementVisitor;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.*;
+import com.intellij.psi.impl.source.tree.LeafPsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.jetbrains.php.lang.psi.PhpPsiElementFactory;
 import com.jetbrains.php.lang.psi.elements.*;
 import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpElementVisitor;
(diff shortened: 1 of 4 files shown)


kalessil:
This is next one for a merge, the merge is coming this week.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java
```diff
@@ -1,19 +1,35 @@
...
+ * (c) Vladimir Reznichenko <kalessil@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
 public class CallableInLoopTerminationConditionInspector extends BasePhpInspection {
-    private static final String messagePattern  = "'for (%existingInit%%newInit%; %newCheck%; ...)' should be used for better performance.";
-    private static final String messageExternal = "Callable result should be stored outside of the loop for better performance.";
+    private static final String message = "Avoid callables in loop conditionals for better performance.";
```
kalessil: @rentalhost: that's less specific reporting as it was before, is it easier to understand due to introducing QF?

src/test/resources/fixtures/semanticalAnalysis/callable-inloop-condition.fixed.php
```diff
@@ -0,0 +1,16 @@
+<?php
```
kalessil: Awesome, thanks for great UTs!

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/CallableInLoopTerminationConditionInspector.java
rentalhost: It avoid tooltips with codes. making it more clear to understand, while QF do the job.

kalessil: Ok, I'll give it a test-drive with the generalized message

Issue #294, opened 2017-05-01
Disallow protected on final classes


rentalhost:
I guess that it doesn't makes sense if you are not overring protected properties or methods.

```php
final class MyClass {
    protected $property;
}
```

It should warning "_$property could be private_".

kalessil:
Nice finding, picking it up.

rentalhost:
Fixed on #314, waiting for merging.

PR #314, opened 2017-05-17
[#294] Disallow protected on final classes

rentalhost:
Following #294. 
_It's not easy-pick as I initially thought, but I think that I did it._ 

It should warn if you declare a class with the `final` modifier and own properties, constants or methods with `protected` modifier. It is non-sense here because `final` classes could not be extended, then the `protected` will acts exactly as `private`.

* [x] Base implementation;
  * [x] Properties support;
  * [x] Constants support;
  * [x] Methods support;
  * [x] Respect super classes implementation;
* [x] UTs implementation;
* [x] QFTs implementation;
* [x] Full code coverage;
* [ ] Text review (_someone?_);


rentalhost:
Fixed.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java
```diff
@@ -0,0 +1,139 @@
...
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder problemsHolder, final boolean isOnTheFly) {
+        return new BasePhpElementVisitor() {
+            private void checkElement(final PhpElementWithModifier element) {
+                if (!element.getModifier().isProtected()) {
+                    return;
+                }
+
+                final PhpClass elementClass = ((PhpClassMember) element).getContainingClass();
+
+                assert elementClass != null;
```
kalessil: Please don't use assert in the plugin codebase, do return if matches null instead

rentalhost: Asserts are good option when we have sure that the value will never fail. For instance, in this assert point we have sure that the class exists, because it will be triggered only from field or method visitor (that are needly a class member). It avoid the NPE warning from IDE and the "return if null". The compiled version will drop the assert, then obvious checks will not be done at runtime. It's used a lot on the intellij API too. But, if you still prefer, I can replace the asserts.

kalessil: Yes, please replace.

EAPs launched with assertions enabled and I don't want the plugin crashed for any reasons.

rentalhost: Done!

rentalhost:
About the assertions, [I used that as reference](http://stackoverflow.com/a/271874/755393) (and comments) and too I have created a [more specific topic](http://stackoverflow.com/questions/44047540/assert-vs-if-null-controlled-case) (currently without replies).

It have some advantages versus the "if null" when we have absolutely sure that it will never fail, because when you are not using `-ea` then this checks will not affects the performance.

Using the [previous commit](https://github.com/rentalhost/phpinspectionsea/blob/ea0cfd4381d90c8356f56978c015b1b3347bf4bf/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java) as reference, I put 4 assertions:

* [L42](https://github.com/rentalhost/phpinspectionsea/blob/ea0cfd4381d90c8356f56978c015b1b3347bf4bf/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java#L42): in this line, we check if the field/method class exists. But it obviously does, because this method is called from a field/method visitor that will be triggered only from some class.
* [L83](https://github.com/rentalhost/phpinspectionsea/blob/ea0cfd4381d90c8356f56978c015b1b3347bf4bf/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java#L83): it always will return something, even if it is a empty array, but never null.
* [L133](https://github.com/rentalhost/phpinspectionsea/blob/ea0cfd4381d90c8356f56978c015b1b3347bf4bf/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java#L133): it always will have a value because it was passed to the constructor of `TheLocalFix` (as `@NotNull`).
* [L134](https://github.com/rentalhost/phpinspectionsea/blob/ea0cfd4381d90c8356f56978c015b1b3347bf4bf/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java#L134): it always will have a value because we just have created it in runtime.

The unique way to it fail is if the IDEA API fails (then it is not our fault). And if we just "ignore" all possible NPE with null comparison, we are just not taking care about test the code it better (basically, any possible error will be supressed).

Just to we think about.

Ps. not all cases should be replaced by assertions, on cases where we have sure that it will pass. Cases where reality is possible get `null` as response still should be done by comparsion.

kalessil:
WIP: merging/test-driving

RULES.md
```diff
@@ -125,6 +125,7 @@ Inspections Lists (Language level migration)
 | Language level migration | PowerOperatorCanBeUsedInspection                | Power operator can be used                          | yes | yes | no   | no  |
 | Language level migration | ConstantCanBeUsedInspection                     | A constant can be used                              | yes | yes | yes  | no  |
 | Language level migration | ArgumentUnpackingCanBeUsedInspection            | Argument unpacking can be used                      | yes | yes | no   | no  |
+| Language level migration | ProtectedWithFinalClassInspector                | Protected modifier could be replaced by private     | no  | yes | no   | no  |
```
kalessil: I think it should be Code Style group, like UnnecessaryFinalModifierInspection.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java
```diff
@@ -0,0 +1,142 @@
...
+import com.intellij.psi.impl.source.tree.LeafPsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.jetbrains.php.lang.psi.PhpPsiElementFactory;
+import com.jetbrains.php.lang.psi.elements.*;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpElementVisitor;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpInspection;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
```
kalessil: Copyright + your name please.

src/test/java/com/kalessil/phpStorm/phpInspectionsEA/languageConstructions/ProtectedWithFinalClassInspectorTest.java
```diff
@@ -0,0 +1,20 @@
+package com.kalessil.phpStorm.phpInspectionsEA.languageConstructions;
+
+import com.intellij.codeInsight.intention.IntentionAction;
+import com.intellij.testFramework.fixtures.CodeInsightFixtureTestCase;
+import com.kalessil.phpStorm.phpInspectionsEA.inspectors.languageConstructions.ProtectedWithFinalClassInspector;
+
+public class ProtectedWithFinalClassInspectorTest extends CodeInsightFixtureTestCase {
+    public void testIfFindsAllPatterns() {
+        myFixture.configureByFile("fixtures/codeSmell/protected-with-final-class.php");
+        myFixture.enableInspections(ProtectedWithFinalClassInspector.class);
```
kalessil: I started cleaning up tests, here should be new ProtectedWithFinalClassInspector() to not introduce warnings.

src/test/resources/fixtures/codeSmell/protected-with-final-class.php
```diff
@@ -0,0 +1,79 @@
...
+    <weak_warning descr="Protected modifier could be replaced by private.">protected</weak_warning> function method2() { }
+    private function method3() { }
+
+    static function method0s() { }
+    static public function method1s() { }
+    static <weak_warning descr="Protected modifier could be replaced by private.">protected</weak_warning> function method2s() { }
+    static private function method3s() { }
+}
+
+class ConstClass1 {
```
kalessil: Any chance to give it more descriptive names?

kalessil:
FYI, I want to merge this PR first \o/, so ping me back once changes have been applied please =)

RULES.md
rentalhost: I can move that to Code Style, but I don't agree with this class name because it treats the "protected" keyword used inside "final" classes (then, "protected with final class"), and  your suggestion seems operates over the "final" keyword ("Unnecessary final modifier"). Can you rethink about that and tell me if it is reality what we should do?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/ProtectedWithFinalClassInspector.java
rentalhost: Fixed.

src/test/java/com/kalessil/phpStorm/phpInspectionsEA/languageConstructions/ProtectedWithFinalClassInspectorTest.java
rentalhost: Fixed.

src/test/resources/fixtures/codeSmell/protected-with-final-class.php
rentalhost: Fixed.

kalessil:
Thank you @rentalhost.

PS: I'll start test-driving the inspection this week.

rentalhost:
And about nightly builds? Should be great if we can accept merges fastly then use nightly builds and we can test-drive all new features side-by-side, and revert merges if it causes some problem (or fixes that).

kalessil:
I started investigating https://docs.travis-ci.com/user/cron-jobs/, seems doable but I need to find out how to publish the build artifacts into beta updates channel (@funivan gave me some hints this year).

kalessil:
I also created a ticket with research results

Issue #255, opened 2017-04-17
Declare type: simplify FQN automatically


rentalhost:
Following #28, is important simplify automatically yet-imported classes.

**Code before declare type:**

```php
use \MyNamespace\AnotherClass;

    ...
    // Inspection: \MyNamespace\AnotherClass can be declared as return type hint.
    /** @return AnotherClass */
    public function getAnotherClass() {
    }
    ...
```

**After QF:**

```php
    ...
    /** @return AnotherClass */
    public function getAnotherClass(): \MyNamespace\AnotherClass {
    }
    ...
```

**Expected after QF:**

```php
    ...
    /** @return AnotherClass */
    public function getAnotherClass(): AnotherClass {
    }
    ...
```

---

**Semi-related**: the tooltip that is shown to current inspection is "': {class} can be...". [There are](https://cloud.githubusercontent.com/assets/421851/25088070/eedda97a-2349-11e7-926b-dcfdc2dd9271.png) a `':` before the class name.

funivan:
It is interesting if we can check phpstorm options. In phpstorm we have option: `automatically import classes`

kalessil:
Should be the same behaviour as for ::class constant usage QF, hence removing easy to pick.

Garethp:
@kalessil I hope you don't mind that I've picked this up. I figured I already knew somewhat how teh `:class constant` QF worked, so it would be a good ticket for me to pick up and dig deeper

kalessil:
Good idea and I have additional input: QF for `::class` and this one should have the same import settings (auto-import, relative FQNs).

The part in `::class` is more stable (and here is rather POC). I'd like the fixer from `::class` to be generalized (without decomposition) for using also in return type hints.

Garethp:
That's the plan. I'll probably end up refactoring the `::class` inspection and QF entirely in the process, since 90% of the logic in there basically relates to what should be common functionality.

Can you please expand on what you mean by
> without decomposition, probably an abstract class with shared importing functionality

My current approach was to create a utility class that contained the following functions: `isClassImported`, `importClass` and possibly `resolveClass` (If the logic from the problem registering would be useful elsewhere). Would you prefer a different approach?

kalessil:
Yes, my idea was doing this:

1) move the fixer into a separate class (we have fixers package) for ::class
2) add an abstract method responsible for the fixing reported expressions (base fixer still big and responsible for importing, but no utils introduced - minimal changes)
[...]

PR #320, opened 2017-05-20
[#255] Suggests the simplified class name on return type declaration

rentalhost:
Following #255 

If you have imported the class `\MyNamespace\MyClass`, then the declared type should be only `MyClass` instead of FQN.

* [x] Declare only the class name if it already imported on file;
  * [x] Respect aliased class name (refer to alias, instead of class name);
* [x] Fixed the tooltip;
* [x] Add support to `Function`;
* [x] Add option to auto-import;

---

~Currently it will not import it automatically. I guess that it should be decided by the user, but maybe I can think about create an option inside this inspection. I will open an issue to work on it after.~


kalessil:
Closing, a follow up ticket has been created: #663

Issue #300, opened 2017-05-06
Code style: fix modifiers ordering


rentalhost:
Tells that `static` should be always on start of a method then the visiblity, or the reverse. For instance:

```php
class {
    // Warning: static should be moved before public.
    public static function () { ... }
}
```

I think that we should have an option list (read-only) to specify what is the prefered order between: abstract, static and visibility (public, private and protected).

Fleshgrinder:
The default PhpStorm is shipped with is `public static function` (or just type `pubsf` and hit tab). Hence, this should be default here, anything else would severely harm productivity, and require people to create new shortcuts.

rentalhost:
It is not a problem... We can consider that as default ordering.

rentalhost:
Closing: fixed on #321

PR #321, opened 2017-05-20
[#300] Missorted modifiers

rentalhost:
Following #300 

This inspection will help sorting the methods modifiers based on [PSR-2: Coding Style Guide](http://www.php-fig.org/psr/psr-2/). 
It declares that:

> [...] `abstract` and `final` MUST be declared before the visibility; `static` MUST be declared after the visibility.

Then it will help ordering the modifiers like:

> [`final` or `abstract`]? [`public `or `protected `or `private`]? [`static`]? `function ()` [...]

---

* [x] Full implementation based on PSR-2 with UTs;
* [x] Support to QF with QFTs;

---

Notes:

1. On the same line is declared too that "_visibility MUST be declared on all properties and methods_". I don't have implemented that because already was implemented on this plugin.
2. Initially I have considered to allow user customize the desired order, but we should follow PSR-2 because it is a standard that is well respected by the community.


kalessil:
PSR2-based: perfect, perhaps naming could reflect PSR - otherwise, we'll get plenty of reports and requests from teams who not following the PSR.

Enabled by default: I have a very bad feeling about it, can you make it optional, please (partially because of the first note)?

kalessil:
Important! Please reassign milestone tickets to me to where merging is pending.

rentalhost:
Okay, some suggestion to how we should rename? I could think about a option to make static come first (non-PSR). I think that we could use that as default with no problem (user can just disable it or customize).

rentalhost:
What I mean about options: create a radio with two options:

- Visibility before "static" (PSR-2 compliance) (default) 
- Visibility after "static"

kalessil:
Naming: what about "PSR-compliant modifiers order"? Does it sound good enough?
No option are needed, as with `{}` people need to follow PSR or just disable the inspection.

rentalhost:
Seems good for me. I will work on this when I come to home.

rentalhost:
Fixed.

kalessil:
@rentalhost: probably this one?

kalessil:
Thank you @rentalhost !

rentalhost:
Great! Seems that the next release will have a lot of good things. :D

kalessil:
Yep, I had to make preview-release as it big one =)

PR #323, opened 2017-05-22
[#293] Detect magic numbers

### LINKED ISSUE #293: Detect Magic Numbers
Could we have support to detect magic numbers? Right now, I'm using this https://github.com/povils/phpmnd to scan my code, but it would be great if this was part of the plugin. the default JS plugin for PHPstorm has support for it already, but there is currently none for PHP.

I'd also mark `0` and `1` as non-magic in a default editable list, which we can add/remove numbers that should be excluded.

rentalhost:
Fixes #293 

It add a lot of analyses to magic numbers, based on [povils/phpmnd](https://github.com/povils/phpmnd) possibilities (except for _arrays_ and _assign_).

**It could** (always enabled):

* **Report for return:** will check for _magic number_ used with the `return` statement in a simple way (eg. `return 0`). Not will check in cases like `return $a = 5`, but we can change that;
* **Report for switch cases:** will check for magic number used as argument for `case` from a `switch` statement (eg. `case 0`);
* **Report for conditionals:** will check for magic number used as operand on a conditional (eg. `$value > 5` or reverse). It will ignore if detects that we are just counting (eg. `count($values) > 0`), but is not too much flexible, then if I do `count($values) > 5`, then `5` is a magic number (_after all, why 5?_).

**Additionally, it could** (_configured by option, but currently enabled by default_):

* **Report on multipler:** will check for magic number as multipler (eg. `$seconds * 60`), except for `-1` because it could be used to change number signal (from positive-to-negative, and reverse);
* **Report on properties:** will check for magic number as default value of properties (eg `public $number = 5`), except if value is `0` (_although I prefer null, in these cases_).
* **Report on parameters:** will check for magic number as default value of parameters (eg. `function ($precision = 8)`), except if value is `0`.
* **Report on arguments:** will check for magic number as argument (eg. `sum($a, 50)`), except if argument is `0` or is the same as the default parameter value (it will follow constant value to identify the resolved number, then it could work together with the previous option without problems).

---

* [x] Implemented with UTs for all cases and options;
* [x] For some reason, options are not been saved. _Some clue?_

Currently without QFs, because it need a more complex process (maybe on future).

### DIFF
--- a/META-INF/plugin.xml
+++ b/META-INF/plugin.xml
@@ -653,6 +653,10 @@ with a laptop.
         shortName="MisorderedModifiersInspection"                 displayName="PSR-compliant modifiers order"
         groupName="Code Style"                                    enabledByDefault="true" level="WEAK WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.MisorderedModifiersInspector"/>
+    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
+        shortName="MagicNumberInspection"                         displayName="Magic number usage"
+        groupName="Code Style"                                    enabledByDefault="true" level="WEAK WARNING"
+        implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.MagicNumberInspector"/>
 
 
     <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
@@ -32,7 +32,6 @@ import com.kalessil.phpStorm.phpInspectionsEA.utils.ElementTypeUtil;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Objects;
 
 import javax.swing.*;
 
@@ -75,7 +74,7 @@ public class MagicNumberInspector extends BasePhpInspection {
                     return;
                 }
 
-                if (Objects.equals(expression.getOperationType(), PhpTokenTypes.opMUL_ASGN)) {
+                if (expression.getOperationType() == PhpTokenTypes.opMUL_ASGN) {
                     final PhpPsiElement expressionValue = expression.getValue();
 
                     if (isNumberOneNegative(expressionValue)) {
@@ -117,7 +116,7 @@ public class MagicNumberInspector extends BasePhpInspection {
 
             @Override
             public void visitPhpBinaryExpression(final BinaryExpression expression) {
-                if (Objects.equals(expression.getOperationType(), PhpTokenTypes.opMUL)) {
+                if (expression.getOperationType() == PhpTokenTypes.opMUL) {
                     if (!optionCheckOnMultiplier) {
                         return;
                     }
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
@@ -4,6 +4,7 @@ import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.codeInspection.ProblemsHolder;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiElementVisitor;
+import com.intellij.psi.PsiReference;
 import com.intellij.psi.tree.IElementType;
 import com.jetbrains.php.lang.lexer.PhpTokenTypes;
 import com.jetbrains.php.lang.parser.PhpElementTypes;
@@ -14,6 +15,9 @@ import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpInspection;
 import com.kalessil.phpStorm.phpInspectionsEA.utils.BinaryExpressionUtil;
 import com.kalessil.phpStorm.phpInspectionsEA.utils.ElementTypeUtil;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Objects;
 
 import org.jetbrains.annotations.NotNull;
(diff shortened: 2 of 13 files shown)


rentalhost:
**My opinion:** I have tried this inspection on a big project that I have. With that I reduced the number of false-positives, by analysing and improving the code. _But, I still having about 167 warnings._ Analysing that I could see that I am making many mistakes and this inspection made this very visible.

For instance:

```php
if (count($innerTeams) < 2) {
    return $innerTeams;
}
```

Reading it fast, I don't know why it is compared by _less than 2_. Certainly a magic number could help me with that.

But I too see a lot of cases that seems to be a false-positive. But seeing closely, you can notice a compliance problem. For instance:

```php
return snake_case(lcfirst(substr($applicationModuleMethod->getName(), 6)), '-');
```

In this case, the `6` argument is applied to `substr()` to remove first 6 characters from string. But if I don't read the code above (_on my project_), I can't understand why it is done.

Then, finally, I guess that this inspection is really very useful. Now we just think about possible false-positives. About NPE I guess that we have none, because I tried on `vendor` and getting nothing (_just one statistic: found over 10.000 cases_).

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
```diff
@@ -0,0 +1,290 @@
...
+import java.util.List;
+import java.util.Objects;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class MagicNumberInspector extends BasePhpInspection {
+    private static final String message = "Magic number should be replaced by a constant.";
+
+    boolean optionCheckOnMultiplier = true;
```
kalessil: public, then PS will start saving them.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
```diff
@@ -0,0 +1,290 @@
...
+                final PsiElement returnArgument = returnStatement.getArgument();
+
+                if (isNumeric(returnArgument)) {
+                    registerProblem(returnArgument);
+                }
+            }
+
+            @Override
+            public void visitPhpBinaryExpression(final BinaryExpression expression) {
+                if (Objects.equals(expression.getOperationType(), PhpTokenTypes.opMUL)) {
```
kalessil: `==` please - the codes are enums/constants alredy

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/binaryOperations/strategy/IdenticalOperandsStrategy.java
```diff
@@ -25,28 +22,14 @@
...
     public static boolean apply(@NotNull BinaryExpression expression, @NotNull ProblemsHolder holder) {
         final PsiElement left  = ExpressionSemanticUtil.getExpressionTroughParenthesis(expression.getLeftOperand());
         final PsiElement right = ExpressionSemanticUtil.getExpressionTroughParenthesis(expression.getRightOperand());
         if (null == left || null == right) {
             return false;
         }
 
-        final IElementType operation = expression.getOperationType();
-        if (operationsToCheck.contains(operation) && PsiEquivalenceUtil.areElementsEquivalent(left, right)) {
+        if (BinaryExpressionUtil.isComparison(expression) && PsiEquivalenceUtil.areElementsEquivalent(left, right)) {
```
kalessil: Please revert + create a separate ticket for internal milestone for migration - I'll do the same for pending refactoring topics this week.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/utils/BinaryExpressionUtil.java
```diff
@@ -0,0 +1,32 @@
...
+import com.jetbrains.php.lang.lexer.PhpTokenTypes;
+import com.jetbrains.php.lang.psi.elements.BinaryExpression;
+
+import java.util.Collection;
+import java.util.HashSet;
+
+import org.jetbrains.annotations.NotNull;
+
+public enum BinaryExpressionUtil {
+    ;
```
kalessil: Is it needed?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
```diff
@@ -0,0 +1,290 @@
...
+                       isBinaryNumeric(numericOperand.getText());
+
+            }
+
+            private boolean isBinaryNumeric(@NotNull final String numericValue) {
+                return "0".equals(numericValue) ||
+                       "1".equals(numericValue);
+            }
+
+            private void registerProblem(final PsiElement rightOperand) {
```
kalessil: No need, inline in-places please.

META-INF/plugin.xml
```diff
@@ -635,6 +635,10 @@ We keep delivering funded results =)
         shortName="UsageOfSilenceOperatorInspection"              displayName="Usage of the silence operator"
         groupName="Code Style"                                    enabledByDefault="true" level="WARNING"
         implementationClass="com.kalessil.phpStorm.phpInspectionsEA.inspectors.codeSmell.UsageOfSilenceOperatorInspector"/>
+    <localInspection language="PHP" groupPath="PHP,Php Inspections (EA Extended)"
+        shortName="MagicNumberInspection"                         displayName="Magic number usage"
+        groupName="Code Style"                                    enabledByDefault="true" level="WEAK WARNING"
```
kalessil: What about magic numbers as 86400, 3600, 60, 24, 1970 - should we consider them as well? I'm asking because you probably gave the inspection test run for some weeks.

And I need you advise if I could pick up another PR instead, which is smaller?

kalessil:
I just started the review, and only later read the comment )
Can you point to a smaller PR - this one will need context awareness and I's eats a lot time....

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
rentalhost: Ahhh... It does make a lot of sense! haha

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
rentalhost: It was a IDE advice (I have enabled all and now I am filtering some haha). But it in this case we have no problem, it will just calls the `equals()` from operation type (that probabily uses directly the `==`). But I will change that.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
rentalhost: I did that to centralize the responsability at one method, once that we have a lot of calls like that. I will change too.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/semanticalAnalysis/binaryOperations/strategy/IdenticalOperandsStrategy.java
rentalhost: I will reply that on BinaryExpressionUtil comment.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/utils/BinaryExpressionUtil.java
rentalhost: I copied the code from L18~26 from another code, then it was marked as "duplicated". Then I refactored it to make it clean and objective. The file that I have copied is fully compatible with that, but I don't modified to not mix issues, but I give the "door open" to make it possible.

META-INF/plugin.xml
rentalhost: All of them are magic numbers (it mean time-related values). Then if you are having problems like that, you should refactoring your code to centralize this kind of value in an unique class. For instance:

```php
class TimeConstants {
    public const HOURS_BY_DAY = 24;
    public const MINUTES_BY_HOUR = 60;
    public const SECONDS_BY_MINUTE = 60;
    // So on...
}
```

And I do not understand your last question (same for your last comment ("I just started the review...")). You mean that you will try another PR before this? If do, no problem, do that.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
rentalhost: Done.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/codeSmell/MagicNumberInspector.java
rentalhost: Done (locally).

rentalhost:
I update the code with your observations and rebased to master. Please, note that I don't applied totally because I have replied and I want to know if you really wants that I do everything. If yes, I will do without problem, else is done. 😄

kalessil:
It's pretty hard decision, but I want to not include magic numbers checking right now. I gave a try to this for my Java projects at the workplace for 2 months - annoys me personally and brought no benefits so far.

10 times sorry @rentalhost, let me send you some merch for this (send me email with your address ad I'll send the stuff after my vacation) =)

rentalhost:
@kalessil Don't care about that! It was fun working on it, and I'm aware that not everything will be accepted. Some features may look awesome on paper but lousy in practice. And to be honest, I deactivated this Java inspection, but I do not remember when or why. hehe

And I don't know what mean "merch", but if this is an _"sorry" materialized in an object_, I will refuse friendly. First of all because I see no reason to be excused at this case, and soo that I live in Brazil and anything you send will only come to here for 3 years or more. haha Thanks anyway!

PR #353, opened 2017-06-13
Add direct links to Acknowledged Projects

bcremer:
Currently the Project Images are linking to the source of the image which is not that useful. This commit adds links to the actual projects's website.


kalessil:
Thank you @bcremer

bcremer:
That was fast :+1:

Issue #361, opened 2017-06-22
Date interval specification validity - report "PT"


downace:
"Probable bugs | Date interval specification validity"

Inspection doesn't report the "PT" string, but it is invalid

![image](https://user-images.githubusercontent.com/7350510/27413797-8dd5d742-5738-11e7-873a-3f2e414ed07f.png)

rentalhost:
**Closing reason: fixed.**
It should be available on next release.

If you have any new question on this issue, please, feel free to open it again.

PR #362, opened 2017-06-22
Date interval specification validity: report "PT" as invalid

rentalhost:
Fixing #361 

* [x] Bug fixed with UTs;
* [x] Code style updated;


kalessil:
Thank you @rentalhost

PR #385, opened 2017-07-08
Add some additional dev packages to the securityAdvisories check

veewee:
I've added 2 more packages which should be required as dev dependency. Couldn't think of more commonly used packages at the moment. 
I've did a quick scroll through https://github.com/exakat/php-static-analysis-tools, you might find one or two more.


kalessil:
Thank you @veewee 

I knew I missed something important there =) Will go thru the list + we plan to add user-level for the settings.

PR #404, opened 2017-07-16
Fix exception with no arguments message grammar

klaude:
This change may make the "thrown exception with no arguments" message more readable to native English speakers.

On a side note, telling users to "do yourself a favor" can be interpreted as condescending. It may be more helpful to explain the benefits of an exception message. Something like "This exception is thrown without a message. Consider adding one to help clarify or troubleshoot the exception." could work. I can put that into this branch too if you like. Thanks for your time!


kalessil:
Thanks @klaude, good to know =) And yes, I think it makes sense to do as you suggested.

klaude:
Thanks, @kalessil! I've updated the message. Squashing it with my previous commit added commit 390412a52ea4783023b63f294b1f4ddecc5a6a39 to the branch. I don't think it'll cause any problems, but if it doesn't work for you I can send my changes in another PR. Thanks again!

kalessil:
Looks good, thank you @klaude !

Issue #501, opened 2017-09-03
Determine const visibility


rentalhost:
It should works like methods, currently.

> PHP | Php Inspections (EA Extended) | Code Style | **Access modifiers shall be defined**

```php
// Warning: "TEST" should be declared with access modifier. 
// Quick-fix: Declare public.
const TEST = true;
```

rentalhost:
WIP

rentalhost:
**Closing:** PR #502

donquixote:
It is nice to see this being supported now.

However, I think there should be a checkbox to disable this inspection for const.

I created a follow-up here: #699

PR #502, opened 2017-09-03
[#501] Check for const visibility

rentalhost:
**Fixing** #501

Instead of create a new inspection for it, I have improved the `AccessModifierPresentedInspector` because it already treats this issue. It just has need some new _patches_.

Note: `const` without modifier is a big problem for PS, once that it don't have a modifiers list if it doesn't exists (eg. `public const` does, but just `const` don't works). In this case, I have implemented a different logic for quick-fix to constants only.

Tests ok and new tests added for PHP 7.1+.

* **Main improvement:**
  * 1fc4d1e951567fcbba74e5a60770a51009b31b66 Added support to visibility check for `const` for PHP 7.1+.
* **General improvement:**
  * 6505c0959fa8abcb7d1ea1de0de284f3c32756ae Quick-fix should respect the PSR-2 modifiers sorting after being applied (as in inspection "_PSR-compliant modifiers order_"). It too give a better "replacement schema" because avoid the regular-expression replacement and the `var` modifier workaround;
  * dd09e16d031b7ebe3591e0eec72242e77c8ecf2f Using `String.format()` instead of regular-expression replacement (turn it more clear and "universal coded");
  * 876caf7d49a5848af1e541ad151d2c3899b39180 Dropping PHT descriptor (`ProblemHighlightType`) to it be handled directly from `plugin.xml` (_please, double check it_);
* **Tests:**
  * d152ac811a84326dd9c77819ffe0b1c0caa884aa Added quick-fixed file (`fixed.php`);
* **Not important:**
  * 6daf8c96202e6b209b6aee02cc82681089da5254 Just updating code-style;
  * 56c82882deba38fb46d575e601db820ce2525c36 Just avoid a variable called `modifiers` that conflicts with class field `modifiers`, and making it more clear: the class field is not `modifiersReference` (once that it is really a reference from real `modifiers` instance).


kalessil:
Oh, good enhancement. Will merge after the 2.3.10 is released.

kalessil:
Thank you @rentalhost

PR #552, opened 2017-09-22
Add Drupal debug functions: dpm/dsm, dvm, kpr and dpq to ForgottenDebugOutputInspector

Ayesh:
Hi Vladimir,
Thanks a lot for this plugin - it simply takes my code to a whole new level!

I work in Drupal projects quite a lot, and use Drupal-specific debug functions (similar to Laravel `dd`) quite a lot. It would be great if they could be marked as Forgotten Debug Output warnings too. I understand one can easily add new functions, but I think they can benefit everyone who uses Drupal with PHPStorm. 

They are global functions (yuck), so I totally understand if you don't feel like adding them to the list and reject this PR. 

Function reference: https://api.drupal.org/api/devel/devel.module/7.x-1.x


kalessil:
Hi @Ayesh, thank you for the PR =) 
We just can not cover all frameworks/CMS and your PR is welcome, so thank you for taking care about Drupal =)

PR #553, opened 2017-09-23
Added debug statement of typo3

nlubisch:


kalessil:
Oh, I tally forgot to check Typo3 docs. Thank you @nlubisch =)

Issue #639, opened 2017-11-15
_<magic method> - missing underscore


kalessil:

zf2timo:
I am interested in contributing this feature. I think it is feasible for a beginner.

kalessil:
Hi Timo, why not =)

1) Dev-environment: https://github.com/kalessil/phpinspectionsea/blob/master/docs/DEVELOPERS.md
2) Target code: https://github.com/kalessil/phpinspectionsea/tree/master/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/magicMethods (please create a separate strategy-class for the feature)

zf2timo:
Thank you. That helps a lot.

I will have a detailed look this evening.

zf2timo:
@kalessil 
When I executed the Unit Test, all tests are failing.
I think It is beacause I am using the JetBrains Toolbox
> Don't use the suggested directory containing app path since it does not contain all needed dependencies.

But how I can find the correct directory?

kalessil:
How does failures message look like?

zf2timo:
As an Example from the `MagicMethodsValidityInspectorTest::testIfFindsInconsistentGetsSetsPatterns` test
>java.lang.NullPointerException
	at com.intellij.openapi.application.impl.ApplicationImpl.isDispatchThread(ApplicationImpl.java:376)
	at com.intellij.testFramework.EdtTestUtilKt.runInEdtAndWait(EdtTestUtil.kt:38)
	at com.intellij.testFramework.EdtTestUtil$Companion.runInEdtAndWait(EdtTestUtil.kt:29)
	at com.intellij.testFramework.EdtTestUtil.runInEdtAndWait(EdtTestUtil.kt)
	at com.intellij.testFramework.UsefulTestCase.runBare(UsefulTestCase.java:413)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:129)
	at junit.framework.TestSuite.runTest(TestSuite.java:252)
	at junit.framework.TestSuite.run(TestSuite.java:247)
	at org.junit.internal.runners.JUnit38ClassRunner.run(JUnit38ClassRunner.java:86)
	at org.junit.runners.Suite.runChild(Suite.java:128)
[...]

PR #648, opened 2017-11-19
Missing Underscore for Magic Methods

zf2timo:
This PR adds a new strategy to detect PHP magic Methods, where a single underscore is used. 
The reporting level is weak Warning.

Solves #639


kalessil:
Thank you @zf2timo

PR #755, opened 2018-01-17
Experimental docker headless PHPStorm configuration

netformic:
Please see the README.md. Works for me :-)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,29 @@
+FROM php:7.0
+
+ENV PHPSTORMVERSION=2017.3 \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM php:7.0
```
afoeder: does/would 7.2 work as well?

didi-at-work: Just tested it. Works flawlessly :-D
I even got additional "Namespace is not bound" warnings that were not reported with 7.0...

kalessil:
Thank you very much @didi-at-work !

PR #798, opened 2018-02-12
Fix stale config

bkeepers:
`closeComment` must be a string, like `markComment`, otherwise it causes an error and issues will never be closed.


kalessil:
Got it, thank you very much @bkeepers.

PR #801, opened 2018-02-14
Fixes #800: cannot iterate over class array

### LINKED ISSUE #800: Cannot iterate (must implement...) on @phpdoc class[]
Problem is similar to "Incorrect inspection error with foreach over iterable  #793" but I'm not sure it same

```
namespace B;
// any class. doesn't reproduce with scalars
/**
 * @param \stdClass[] $array
 */
function aaa(array $array)
{
   //  array_reverse got false positive error 
    foreach (array_reverse($array) as $item) {
        $item->a();
    }
}
```
phpstorm

rentalhost:
* 1d2ff48: bugfix;
  * Actually it was just a simple `if()` for negate the `types.contains(Types.strArray)`;
* 2b57488: updated the code-style;
  * L155: Fixed non-null argument;
  * `String.replace()` was changed to `String.format()`;

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
@@ -150,7 +150,6 @@ public class ForeachSourceInspector extends BasePhpInspection {
                     }
                 }
 
-
                 /* gracefully request to specify exact types which can appear (mixed, object) */
                 if (types.contains(Types.strMixed)) {
                     /* false-positive: mixed definitions from stub functions */
@@ -181,8 +180,12 @@ public class ForeachSourceInspector extends BasePhpInspection {
                     types.remove(Types.strNull);
                 }
 
+                /* if contains an array, then it is iterable */
+                if (types.contains(Types.strArray)) {
+                    return;
+                }
+
                 /* do not process foreach-compatible types */
-                types.remove(Types.strArray);
                 types.remove(Types.strIterable);
                 types.remove("\\Traversable");
                 types.remove("\\Iterator");
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
@@ -40,9 +40,9 @@ public class ForeachSourceInspector extends BasePhpInspection {
     public boolean REPORT_UNRECOGNIZED_TYPES = true;
 
     private static final String patternNotRecognized = "Expressions' type was not recognized, please check type hints.";
-    private static final String patternMixedTypes    = "Expressions' type contains '%t%', please specify possible types instead (best practices).";
-    private static final String patternScalar        = "Can not iterate '%t%' (re-check type hints).";
-    private static final String patternObject        = "Iterates over '%t%' properties (probably should implement one of Iterator interfaces).";
+    private static final String patternMixedTypes    = "Expressions' type contains '%s', please specify possible types instead (best practices).";
+    private static final String patternScalar        = "Can not iterate '%s' (re-check type hints).";
+    private static final String patternObject        = "Iterates over '%s' properties (probably should implement one of Iterator interfaces).";
 
     @NotNull
     public String getShortName() {
@@ -57,7 +57,7 @@ public class ForeachSourceInspector extends BasePhpInspection {
             public void visitPhpForeach(@NotNull ForeachStatement foreach) {
                 final PsiElement source = ExpressionSemanticUtil.getExpressionTroughParenthesis(foreach.getArray());
                 if (source instanceof PhpTypedElement && !isEnsuredByPyParentIf(foreach, source)) {
-                    this.analyseContainer(source);
+                    analyseContainer(source);
                 }
             }
 
@@ -98,7 +98,7 @@ public class ForeachSourceInspector extends BasePhpInspection {
                     return;
                 }
                 final Set<String> types = resolvedType.filterUnknown().getTypes().stream()
-                        .map(Types::getType).collect(Collectors.toSet());
+                                                      .map(Types::getType).collect(Collectors.toSet());
                 if (types.isEmpty()) {
                     /* false-positives: pre-defined variables */
                     if (container instanceof Variable) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
```diff
@@ -92,60 +113,76 @@ private boolean isEnsuredByPyParentIf(@NotNull ForeachStatement foreach, @NotNul
...
                     }
+
                     return;
                 }
 
                 /* false-positives: multiple return types checked only in function/method; no global context */
                 final PsiElement scope = ExpressionSemanticUtil.getBlockScope(container);
-                if (types.size() > 1 && !(scope instanceof Function)) {
+
+                if ((types.size() > 1) && !(scope instanceof Function)) {
```
kalessil: Where the extra line + extra `()` are coming from? 
It would make my reviewer life easier to get the original code style back =)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
```diff
@@ -92,60 +113,76 @@ private boolean isEnsuredByPyParentIf(@NotNull ForeachStatement foreach, @NotNul
                 return result;
             }
 
-            private void analyseContainer(@NotNull PsiElement container) {
+            private void analyseContainer(@NotNull final PsiElement container) {
```
kalessil: Parameters don't need to be final, only if has a purpose (but local variables are final for sure).

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
```diff
@@ -54,33 +73,35 @@ public String getShortName() {
...
-                    if (ifCandidate instanceof If) {
-                        conditions = ((If) ifCandidate).getCondition();
-                    } else if (ifCandidate instanceof ElseIf) {
-                        conditions = ((ElseIf) ifCandidate).getCondition();
-                    } else {
-                        conditions = null;
+                    final PsiElement ifCandidate = (foreach.getParent() instanceof GroupStatement) ? foreach.getParent().getParent() : null;
+                    PsiElement       conditions  = null;
+
+                    if (ifCandidate instanceof ControlStatement) {
```
kalessil: Catches lots of other constructs (not only If, ElseIf) - is it intended?

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
```diff
@@ -35,15 +54,15 @@
  */
 
 public class ForeachSourceInspector extends BasePhpInspection {
+    private static final String patternNotRecognized = "Expressions' type was not recognized, please check type hints.";
+    private static final String patternMixedTypes    = "Expressions' type contains '%s', please specify possible types instead (best practices).";
+    private static final String patternScalar        = "Can not iterate '%s' (re-check type hints).";
+    private static final String patternObject        = "Can not iterate '%s' (must implement one of Iterator interfaces).";
+
```
kalessil: +1, thank you =)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
rentalhost: By topics:

1. The extra line is just to make clear the `if()` vs variable initialization. The code is more "spaced" and it turn the readability better in "long term usage".
2. The extra `()` is IDE thing actually. I enabled a lot of Java inspections on past and one of them is something like "unclear expression". It requires that we add an extra `()` in complex conditionals to delimit each part of expression. It is useful in sometimes, but cases like that I don't think that is very useful, but I too don't care about that. But I can undo that if you prefer. Maybe disable this inspection for you project.
3. Oh. Don't care too much with "update code-style" commits, once that it don't modify the code (except if I describe it - actually I merged the `String.format()` part with this commit accidentally, because the idea is that this commit fix only the style and not modify behavior). It is basically to make code readability better and solve all IDE inspections.

Anyway, if you don't like it I could make a new PR with only the real fix.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
rentalhost: While it really is not need, I think that is valid to make sure that this parameter is never modified.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
rentalhost: It will catch some additional constructs, really. But will support things like that:

```php
switch (true){
    case is_array($value):
        foreach ($value as $item) {
        }
        break;
}
```

And too to `while()`, `do while()` and `for()` conditionals. But we can limit to works strictly with that types, once that it will includes `catch()` and `finally()` that don't have really a conditional.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
kalessil: Yes, please (keep the fix and String.format changes only) =)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
kalessil: No need ATM

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/forEach/ForeachSourceInspector.java
kalessil: Let's stick to If, ElseIf - I'd like to minimize this inspection modification - it's very problematic one =_

rentalhost:
I'll back to work on this fix on this weekend.

kalessil:
@rentalhost I made changes in master, leading to conflict. Sorry.

rentalhost:
@kalessil No problem! I will rework on this source to make this PR clean.

rentalhost:
Okay, I have reduced all modifications to fixes the `String.replace()` to `String.format()`, converting to regular comparison style where was Yoda style and the fix itself. But Github shows all changes related to space things, so the changes history still is confuses, but less than before. :smile: 

In anyway, I think that this inspection should be rewrite in some moment, once that it is very confuses and seems few optimized. The `analyseContainer()` method for instance is very long and confuses. I don't, for instance, if I can move the fix to top (it don't break any test, but I think that is not a good thing at this time).

kalessil:
Looks good, I'll merge today.

kalessil:
Thank you @rentalhost

Issue #797, opened 2018-02-11
Replace backtick to shell_exec()


rentalhost:
Backticks have the same effect as `shell_exec()`, but it could causes confusion because in some fonts you will see something like a quote.

**Todo:**
* **Inspection**: suggests replace backtick to `shell_exec()`;
* **Quick-fix**: escaped backticks inside backticks should be unescaped;
* **Quick-fix**: unescaped single quotes inside backticks should be escaped;

**References:**
* http://php.net/manual/en/language.operators.execution.php
* https://wiki.php.net/rfc/deprecate-backtick-operator

kalessil:
Oh my, I wasn't even aware of this crap. Goes to language level migration (hopefully PS parser recognizes this construct at all)

PR #802, opened 2018-02-14
Feature #797: avoid backtick operator usage

rentalhost:
* 2dfdc9e: inspection implementation;
  * a1321d9: quick-fix implementation;
* 3e806b9: code-style updated;
  * Methods `isTestContext()` and `isFromRootNamespace()` was modifed to be `static`;

It will converts to `shell_exec("double-quoted")` once that single-quotes will not supports variables inside backticks. In future we can try identify better quote style based on backtick contents.

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/BacktickOperatorUsageInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/BacktickOperatorUsageInspector.java
@@ -1,0 +1,94 @@
+package com.kalessil.phpStorm.phpInspectionsEA.inspectors.languageConstructions;
+
+import com.intellij.codeInspection.LocalQuickFix;
+import com.intellij.codeInspection.ProblemDescriptor;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiElementVisitor;
+import com.intellij.psi.SmartPointerManager;
+import com.intellij.psi.SmartPsiElementPointer;
+import com.jetbrains.php.lang.psi.PhpPsiElementFactory;
+import com.jetbrains.php.lang.psi.elements.ParenthesizedExpression;
+import com.jetbrains.php.lang.psi.elements.PhpShellCommandExpression;
+import com.jetbrains.php.util.PhpStringUtil;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpElementVisitor;
+import com.kalessil.phpStorm.phpInspectionsEA.openApi.BasePhpInspection;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.regex.Pattern;
+
+/*
+ * This file is part of the Php Inspections (EA Extended) package.
+ *
+ * (c) Vladimir Reznichenko <kalessil@gmail.com>
+ * (c) David Rodrigues <david.proweb@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+public class BacktickOperatorUsageInspector extends BasePhpInspection {
+    static final String message = "Prefer shell_exec() function instead of this backtick operator";
+
+    @NotNull
+    public String getShortName() {
+        return "BacktickOperatorUsageInspection";
+    }
+
+    @Override
+    @NotNull
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, final boolean isOnTheFly) {
+        return new BasePhpElementVisitor() {
+            @Override
+            public void visitPhpShellCommand(@NotNull final PhpShellCommandExpression shellCommandExpression) {
+                holder.registerProblem(shellCommandExpression, message, new UseShellExecQuickFix(shellCommandExpression));
+            }
+        };
+    }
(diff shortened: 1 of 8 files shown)


rentalhost:
I will check what happen in travis on this weekend.

kalessil:
Ok, let me know if ran out of ideas =)

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/PackedHashtableOptimizationInspector.java
```diff
@@ -54,7 +54,7 @@ public void visitPhpArrayCreationExpression(ArrayCreationExpression expression)
                 }
 
                 /* false-positives: test classes */
-                if (this.isTestContext(expression)) {
+                if (BasePhpElementVisitor.isTestContext(expression)) {
```
kalessil: I'd prefer to stick avoid this change due to merging into ultimate

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/openApi/BasePhpElementVisitor.java
```diff
@@ -20,86 +22,177 @@
  */
 
 public abstract class BasePhpElementVisitor extends PhpElementVisitor {
+    protected static boolean isTestContext(@NotNull final PsiElement expression) {
```
kalessil: Same here: merging reasons, only essential changes here

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/openApi/BasePhpElementVisitor.java
```diff
@@ -20,86 +22,177 @@
...
+    public void visitPhpEval(@NotNull final PhpEval eval) {
+    }
+
+    public void visitPhpDeclare(@NotNull final Declare declare) {
+    }
+
+    public void visitPhpDocTag(@NotNull final PhpDocTag tag) {
+    }
+
+    public void visitPhpShellCommand(final PhpShellCommandExpression shellCommandExpression) {
```
kalessil: +1

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/BacktickOperatorUsageInspector.java
```diff
@@ -0,0 +1,96 @@
...
+    public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, final boolean isOnTheFly) {
+        return new BasePhpElementVisitor() {
+            @Override
+            public void visitPhpShellCommand(@NotNull final PhpShellCommandExpression shellCommandExpression) {
+                holder.registerProblem(shellCommandExpression, message, new UseShellExecQuickFix(shellCommandExpression));
+            }
+        };
+    }
+
+    private static class UseShellExecQuickFix implements LocalQuickFix {
```
kalessil: There is an easy way for this: com.kalessil.phpStorm.phpInspectionsEA.inspectors.apiUsage.IsNullFunctionUsageInspector.CompareToNullFix. Visitor will generate text for replacement and the fix will apply it (the base fixer is stable and bug-free as used globally).

rentalhost: Oh, great. I will check that way!

rentalhost: I am working on this PR right now and I checked your suggestion here, but it seems to be an excessive work to inspection if the fix is not applied (the inspection will generates the fix, allocate it, but could not be used by the user; and the fix will be regenerated on each code change). Am I right?

rentalhost:
I have no clue about this error. I can't understand what this log mean (actually I cannot see the last log because it shows "_there was an error while trying to fetch the log_"). It works on last version (and EAP), but fails on another versions.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/BacktickOperatorUsageInspector.java
kalessil: Yes, you are right. A little bit of text manipulation and fixer creation is taking an extremely small amount of resources, in fact, I migrating to this approach last year.

kalessil:
That's parser, older versions are not supporting nested back-ticks (both escaped and un-escaped).
I'd suggest limiting test fixture to simple cases.

src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/languageConstructions/BacktickOperatorUsageInspector.java
rentalhost: I think that it could be simplified without use additional resources (even than small) by using closures. Maybe we can implements that or you think better do like you suggested anyway?

rentalhost:
Oh! It is bad, once that is important check if backticks will be unescaped. Well, anyway I think that we have no better solution for now. :(

kalessil:
@rentalhost : logic still there as i understand, just a couple test cases are missing - no worries =)

kalessil:
@rentalhost : merged, thank you!

PR #841, opened 2018-03-06
Probable bugs doc review.

mikemilano:
Below was the only sentence I didn't quite understand enough to work with:

> PHP´s magic [`::class`-constant](http://php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class) will not 
> canonical the casing of your imports.

### DIFF
--- a/docs/probable-bugs.md
+++ b/docs/probable-bugs.md
@@ -3,11 +3,12 @@ This document provides help on getting your code bug free.
 
 ## Phar-incompatible 'realpath()' usage
 
-The realpath() expands all symbolic links and resolves references to '/./', '/../' and extra '/' characters in 
-the input path and returns the canonical absolute pathname. But what about streams, like phar://<file-path>?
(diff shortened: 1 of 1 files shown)


bcremer:
Please see my gist here for a example regarding the `::class` problematic: https://gist.github.com/bcremer/9e8d6903ae38a25784fb1985967c6056#file-php-class-name-constant-case-sensitivity-and-psr-11-md

mikemilano:
@bcremer it's unclear what I'm looking for in your gist. To be clear, the example made sense, it was just the wording of the sentence I wasn't clear on the intended meaning. As I understand it, canonical is an adjective but it's being used as a verb.

bcremer:
Ah I see. Should be "canonize" instead of "canonical" in this sentence. I'm not sure if that is the best word after all. I'm not native english speaker.

kalessil:
Many thanks @mikemilano =)

PR #1005, opened 2018-08-29
Migrate to gradle based build

cedricziel:
This change introduces Gradle as the build system. The purpose is to make it
easier to build, with less manual steps.

Basically new contributors need to:

* install the JDK
* clone the repo
* use IntelliJ (CE is sufficient) and import the `build.gradle` file as
  project from existing sources file, check "auto-import" and
  "auto create source roots", use the included wrapper as recommended
* nothing. That's it

The project leverages the Gradle wrapper which makes the individual
projects' build-tool independent from a (possibily) globally installed
gradle version.

All manual build commands should begin with `./gradlew` (windows users
should use the `.bat` file.

### Available tasks:

* `./gradlew buildPlugin` - builds a zip file in `.gradle/distributions`
* `./gradlew check` - runs the tests
* `./gradlew runIDE` - launches the currently configured IDE version
  with the plugin and dependencies installed

## Versioning

Currently, the version is derived from the git tag, and `plugin.xml` is
automatically through the build script. The current configuration
modifies the `since` and `until` properties to match the
build-environment.

Of course it's possible to change, but it's also possible to
automatically publish the plugin (https://github.com/JetBrains/gradle-intellij-plugin#publishing-dsl,
for reference: https://github.com/cedricziel/idea-php-typo3-plugin:.travis.yml@master#L36:L53), so that's perfectly fine, but also changeable
if you need to publish it manually.

## Building for different IDEA versions

The configuration relies on gradle properties to determine the current
IDEA and PHP Plugin version.

To switch the current version, comment out the current config in
`gradle.properties` and un-comment the desired version. IDEA will pickup
the change  and reconfigure the project. In travis it's mostly the same,
but environment variable-driven.

## Running and debugging

Use IDEA to execute the `runIde` gradle task to run a IDE. Use the same
task as debug-run and set your breakpoints. Use the `reload classes`
task to recompile and hot-reload.

When reloading fails, you might want to delete the `out` folder and / or
the `.gradle/classes` directory and reload again.


.travis.yml
```diff
@@ -10,26 +10,46 @@ addons:
...
 cache:
   directories:
-    - .travis/.cache
+  - $HOME/.gradle/caches/
+  - $HOME/.gradle/wrapper/
+
+before_install:
+- "export ORG_GRADLE_PROJECT_ideaVersion=${IDEA_VERSION}"
+- "export ORG_GRADLE_PROJECT_phpPluginVersion=${PHP_PLUGIN_VERSION}"
+- "export ORG_GRADLE_PROJECT_psiViewerPluginVersion=${PSI_VIEWER_PLUGIN_VERSION}"
```
funivan: You can skip this line. All versions of PSI_Plugin are the same. So you do not need to specify (IMHO)

cedricziel: We certainly could - somehow. I'm not aware on how to cascade dependencies for different environments.

If you're asking me - I'd just keep it since it's small, convenient and not embedded in the final artifacts.

cedricziel: Yes, I would keep it. Assembling the array of dependencies out of band makes it just harder to read and nothing is gained from it. - IMHO.

kalessil: Let's keep it, I have to install it all the time locally.

kalessil:
That's just wow, thank you @cedricziel! Will merge shortly.

cedricziel:
I'd appreciate a license, but that's not the main purpose of the PR :)

kalessil:
For sure, will send it today evening ;)

cedricziel:
Thanks a bunch!

kalessil:
Sent! Also I pushed some changes today, would you please resolve the caused conflicts (sorry!)?

cedricziel:
Hey thanks for the license! :)

I just rebased the PR - also be sure to let me know if you need anything.

kalessil:
Thank you @cedricziel, merged!

cedricziel:
So great! :)

funivan:
This is Awesome +1 @cedricziel

Issue #1106, opened 2018-11-16
extending "__" methods from SoapClient class causes a false positive


thomasphansen:
Let's start with common details:
 
Subject: False positive when extending SoapClient 
Details: Php has legacy classes methods that start with "__" (double underscore). If you try to overwrite one of them when extending those classe, you get a false positive error.
| Issue type     | Bug/False-Positive|
| Plugin         | Php Inspections (EA Extended) v3.0.8 |
| Language level | PHP 7.1 |

Current behaviour (description/screenshot:)
extending "__" methods from legacy classes (eg. SoapClient) results in error:
  Only magic methods should start with '__'.

Expected behaviour
no error should be shown when extending legacy classes methods starting with "__"

Information from Help -> About dialog (contains IDE and it's version details).
PhpStorm 2018.3 EAP
Build #PS-183.4284.100, built on November 14, 2018
PhpStorm EAP User
Expiration date: December 14, 2018
JRE: 1.8.0_152-release-1343-b15 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 4.10.0-42-generic

I'm creating a push request to fix this.

kalessil:
Covered by #1107

PR #1107, opened 2018-11-16
Issue #1106 - overwriting methods starting with __ from legacy classes

thomasphansen:
This is my first contribution to a java project, so please be kind! :)

This PR fixes Issue #1106, by creating an array of "legacy" classes that can be tested against the found method. If the current class extends a legacy class and the method has a parent there, then it will not trigger the "Only magic methods should start with '__'" error message.

Hope it helps! :)


kalessil:
Thank you @thomasphansen, merged.

PR #1182, opened 2019-01-24
Changed Path in the entrypoint.sh because the path

filipe96:
strcuture of the new PhpStorm version changed.
In the Dockerfile I changed the php Version and
the PhpStorm Version Number.
On top of this the link to the phpinspectionea-Plugin
is not Valid enymore. So I changed it.


kalessil:
Thank you @filipe96

PR #1315, opened 2019-05-02
Fix typo

Vusys:
Possessive its has no apostrophe


kalessil:
Merged, thank you @Vusys

PR #1392, opened 2019-08-22
Improve "Slow array function used in loop" example readablillity

bcremer:
Replace empty array initialization with empty array during the final array_merge call.

This might be controversial and personnel preference but I find the empty array initialization is not obvious  and can better be moved down to the actual `array_merge` call.

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -30,17 +30,17 @@ Let's start with an example demonstrating the case:
 
 In order to reduce execution time we can modify the code and perform the merge operation only once:
 ```php
-    /* the inner empty array covers cases when no loops were made */
-    $options = [[]];
+    $options = [];
     foreach ($configurationSources as $source) {
         /* something happens here */
         $options[] = $source->getOptions(); // <- yes, we'll use a little bit more memory
     }
+   
     /* PHP below 5.6 */
-    $options = call_user_func_array('array_merge', $options);
+    $options = call_user_func_array('array_merge', [] + $options); // the empty array covers cases when no loops were made
 
     /* PHP 5.6+: more friendly to refactoring as less magic involved */
-    $options = array_merge(...$options);
+    $options = array_merge([], ...$options); // the empty array covers cases when no loops were made
 ```
 
 ## Foreach variables reference usage correctness
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -37,7 +37,7 @@ In order to reduce execution time we can modify the code and perform the merge o
     }
    
     /* PHP below 5.6 */
-    $options = call_user_func_array('array_merge', [] + $options); // the empty array covers cases when no loops were made
+    $options = call_user_func_array('array_merge', $options + [[]]); // the nested empty array covers cases when no loops were made, must be second operand
 
     /* PHP 5.6+: more friendly to refactoring as less magic involved */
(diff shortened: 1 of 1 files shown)


SpyroTEQ:
Maybe I'm missing something (because I'm new here; I do have so inspections in my own personnal plugin and I'm taking a look to see if they would fit in the more general and known PHP EA plugin), but this command line would fail:

```
php -r "var_dump(call_user_func_array('array_merge', [] + []));"

Warning: array_merge() expects at least 1 parameter, 0 given in Command line code on line 1

Call Stack:
    0.0418     392792   1. {main}() Command line code:0
    0.0418     392792   2. array_merge() Command line code:1

Command line code:1:
NULL
```

Hence why arrays were nested at first?

kalessil:
@SpyroTEQ : agree it should be '[[]]' there. Regarding ternaries: they are adding +1 each one to Cyclomatic complexity, we can not promote this way (as we are trying to reduce the complexity as much as possible).

In fact I'm only trying to find more elegant solution for 'call_user_func_array('array_merge', [[]] + $options)'.

docs/performance.md
```diff
@@ -30,17 +30,17 @@ Let's start with an example demonstrating the case:
...
-    $options = [[]];
+    $options = [];
     foreach ($configurationSources as $source) {
         /* something happens here */
         $options[] = $source->getOptions(); // <- yes, we'll use a little bit more memory
     }
+   
     /* PHP below 5.6 */
-    $options = call_user_func_array('array_merge', $options);
+    $options = call_user_func_array('array_merge', [] + $options); // the empty array covers cases when no loops were made
```
kalessil: @bcremer: this will not work as expected: https://3v4l.org/IjHO1, please update to 'call_user_func_array('array_merge', $options + [[]])' (with a comment that placing '[[]]' as the second operand is important) and I'll merge the change right ahead.

docs/performance.md
```diff
@@ -30,17 +30,17 @@ Let's start with an example demonstrating the case:
...
 
     /* PHP 5.6+: more friendly to refactoring as less magic involved */
-    $options = array_merge(...$options);
+    $options = array_merge([], ...$options); // the empty array covers cases when no loops were made
```
kalessil: Nice!

docs/performance.md
kalessil: @bcremer: wrong link, sorry: https://3v4l.org/Jm5Eh

kalessil:
@SpyroTEQ : Oh cool, do you have any public link?

SpyroTEQ:
@kalessil If you're talking about the inspections I have, no, I don't have public link... And I sadly checked afterward and saw that I mainly have SQL inspections and no generic-friendly PHP inspections :/

kalessil:
@SpyroTEQ: yep, I meant your inspections. Understood, hope you'll find the Extended plugin useful then - you'll probably need to configure it for your needs, but afterwards it's a great help.

SpyroTEQ:
@kalessil Be sure I already use it and love it since months : ) And that's probably why I did not need any PHP inspection (and only SQL inspections):; there is no SQL Inspection EA !

docs/performance.md
bcremer: @kalessil Updated the PR. Sorry for not checking the change better beforehand.

kalessil:
Thank you @bcremer, merged!

PR #1401, opened 2019-09-03
Add plugin icon

funivan:
In latest versions of the IDE plugin can have an icon. 
This is how it may look like.


kalessil:
Merged, thanks!

PR #1423, opened 2019-10-03
Remove github since you're not enrolled into sponsorship program

samdark:
image

Since you are not enrolled to GitHub sponsoring program (I'm not aware of a single developer who's enrolled), github is better to be removed from the list.


kalessil:
Oh, thank you @samdark !

PR #1561, opened 2020-05-13
deprecated constructor inspection should not hint static method with same as class name

Haehnchen:
Hinting "deprecated constructor" on the function `Foo` is not validate as i would think

```
class Foo {
public static function Foo() {}
}
```


kalessil:
Thank you @Haehnchen !

PR #1576, opened 2020-05-29
Update testData according to PhpStorm Control Flow engine changes

wbars:
Hi. Today we're plan to release first PhpStorm 2020.2 EAP. In this version simple "always-false" instructions will not be reachable from control flow entry point anymore, since they're considered as unreachable code (as if there was 'return' for instance). This PR changes testData to make test pass, but please feel free to tune inspection code.


kalessil:
Merged, thank you @wbars!

PR #1577, opened 2020-05-29
Resolve class references from PhpTypeDeclaration rather than from first parameter child

wbars:
Hi. Today we're plan to release first PhpStorm 2020.2 EAP. In this version we have to make PSI-breaking changes regarding union types support: in parameters and field types all class references in these types will be wrapped into separate `PhpTypeDeclaration` PSI element, so class references should be now fetched from it rather than from parameter.


kalessil:
Thank you for the PR @wbars and the note regarding upcoming changes ( also greetings to Alex and the team =) ). I'm currently looking into providing BC changes, so CI is not breaking - will merge as git to this point.

kalessil:
@wbars: checked, since PhpTypeDeclaration is not available before PS 2020.2 I had to use PsiTreeUtil and traverse parameter node for finding references (tree materialization is not a problem, it's a tiny psi fragment). Here is my solution: https://github.com/kalessil/phpinspectionsea/commit/94f765131d7d35f0a2d76706ddac75c68a5c3b3e (identical the one from you, just another way of fetching references).

Issue #1628, opened 2020-09-07
False Positive: Php reserved $http_response_header variable is not set


simonberger:
PHP Language level: 5.3 (shouldn't  matter)

$http_response_header is reported with `seems to be not defined in the scope`
But it is a kind of PHP global variable.
https://www.php.net/manual/en/reserved.variables.httpresponseheader.php

rentalhost:
Confirmed.

```php
function test()
{
    isset($http_response_header);
}
```

kalessil:
Fixed in #1629

dryabov:
@kalessil Why other superglobals are resolved properly, and these ones aren't?
And by the way, rarely used `$argc` and `$argv` are not resolved too.

PS. I was sure PhpStorm reads superglobals from the `_superglobals.php` stub file, but now I'm doubtful about it (because it highlights `$argc`, `$argv`, and `$HTTP_RAW_POST_DATA` as undefined).

PR #1629, opened 2020-09-07
Added $http_response_header to the list of reserved variables

simonberger:
Addresses: #1628 

Here is a list of all reserved variables.
https://www.php.net/manual/en/reserved.variables.php

I don't know if any of the other more known variables is affected as well or filtered out differently.


kalessil:
LGTM, thank you @simonberger!

PR #1647, opened 2020-10-20
Update roave/security-advisories requirement

nicolasbeauvais:
Hello,

I updated the [roave/security-advisories ](https://github.com/Roave/SecurityAdvisories) requirement to use `dev-latest` instead of `dev-master` as it has been changed by the maintainers, cf: https://github.com/Roave/SecurityAdvisories/commit/5ae6aa6482b72946bc0d279c3b54dc1dbeae68d8

Cheers.

PR #1682, opened 2021-01-15
Allow dev-latest for roave/security-advisories

pataar:
`dev-latest` should be valid as well. This is stated in their documentation to use. I think it should be available as alternative.

See: https://github.com/Roave/SecurityAdvisories

I know this is pretty similar to #1647, although this would have less impact as existing configurations still work.


ea-inspections-team:
Thank you for the PR @pataar for a great addition. Would you please fix the merge conflict, so we could go ahead with this PR?

PR #1716, opened 2021-04-08
Fix leading and trailing messages

level39:


ea-inspections-team:
Looks good, thank you @level39.

PR #1750, opened 2021-07-21
minor cs-fix: array() -> []

ktomk:
as PHP_VERSION_ID >= 70000 and already [] on that line for the inner.

one whitespace issue on the last line?


ea-inspections-team:
Thank you @ktomk !

PR #1757, opened 2021-08-12
Update ShortEchoTagCanBeUsedInspector.java

cyingfan:
Just removing the incorrect message. Disabling short_open_tag does not disable short echo tag.


ea-inspections-team:
Thank you @cyingfan !

PR #1775, opened 2021-10-19
[phpUnit] Suggest assert{Directory,File}DoesNotExist instead of deprecated assert{DirectoryFile}NotExists

caugner:
See: https://github.com/sebastianbergmann/phpunit/issues/4077


ea-inspections-team:
Thanks for working on the PR @caugner!

It was more work required to handle the deprecations, we took the liberty to address it in https://github.com/kalessil/phpinspectionsea/commit/edf7597546846d279c855a78858ac8acb981fadc and https://github.com/kalessil/phpinspectionsea/commit/bc18000de3e8008e0d78815446cdfdb3ec943cad in the most graceful way.

Issue #1725, opened 2021-05-05
[False positive] JSON_THROW_ON_ERROR is not recognized in named parameter call (PHP 8.0)


niconoe-:
| Subject        | Details                                                                       |
| :------------- | :---------------------------------------------------------------------------- |
| Plugin         | Php Inspections (EA Extended) **4.0.6.3** |
| Language level | PHP 8.0                        |

Current behaviour
--
When calling `json_decode` with the `flags:` names parameter set to JSON_THROW_ON_ERROR, the plugin is rising an error about `[EA] Please consider taking advantage of JSON_THROW_ON_ERROR flag for this call options.`

![image](https://user-images.githubusercontent.com/9560327/117152052-f079c800-adb9-11eb-98e0-9c641c7924d8.png)

Expected behaviour
--
Expected behaviour is "no error" as I am taking advantage of the Throwable behavior.

Environment details
---
<pre>
PhpStorm 2021.1.2
Build #PS-211.7142.44, built on April 30, 2021
Subscription is active until February 22, 2022.
Runtime version: 11.0.10+9-b1341.41 amd64
VM: Dynamic Code Evolution 64-Bit Server VM by JetBrains s.r.o.
Linux 4.15.0-142-generic
GC: G1 Young Generation, G1 Old Generation
Memory: 2048M
Cores: 4
Registry: run.processes.with.pty=TRUE, ide.tooltip.initialDelay=600
Non-Bundled Plugins: au.com.glassechidna.luanalysis (1.2.3), manjaro.mpb (1.5), me.aristotll.python.typing.adder (1.1), mobi.hsz.idea.gitignore (4.1.0), org.plugin.dot.id (1.2), com.kalessil.phpStorm.phpInspectionsEA (4.0.6.3), de.espend.idea.php.annotation (8.0.0), PlantUML integration (5.3.0)
Current Desktop: Unity
</pre>

Probably some actions have been already taken regarding #1555 or #1592 but those were not about PHP 8.0 and the ability to use a named parameter.

Thanks a lot :slightly_smiling_face:

niconoe-:
I don't speak Java and I never developped a PHPStorm pluggin, so I don't know how I could help in fixing this.

What I think is something can be done in this snippet:
[...]

PR #1808, opened 2022-02-07
Write unit tests for name identifiers in json_decode

winkbrace:
Pull request regarding #1725 
I built failing test cases, but unfortunately, like [niconoe-](https://github.com/niconoe-), I'm not a Java developer and can't make the time right now to figure out how to run this project, run the unit tests, and how to get argument names from a PsiElement.

It should be a relatively simple fix to check for an argument name from the `arguments` list in [JsonEncodingApiUsageInspector.java](https://github.com/kalessil/phpinspectionsea/blob/ad57a4a557daff9fa97df654c467ac5048c2379b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/JsonEncodingApiUsageInspector.java#L78-L95).

Oh well, hopefully this is somewhat useful and otherwise feel free to decline.

### DIFF
--- a/src/test/java/com/kalessil/phpStorm/phpInspectionsEA/api/JsonEncodingApiUsageInspectorTest.java
+++ b/src/test/java/com/kalessil/phpStorm/phpInspectionsEA/api/JsonEncodingApiUsageInspectorTest.java
@@ -38,4 +38,21 @@ final public class JsonEncodingApiUsageInspectorTest extends PhpCodeInsightFixtu
             myFixture.checkResultByFile("testData/fixtures/api/json-encoding-errors-handling.fixed.php");
         }
     }
+    public void testItRecognizesNameIdentifierFlags() {
+        final PhpLanguageLevel level = PhpLanguageLevel.parse("8.0");
+        if (level != null && level.getVersionString().equals("8.0")) {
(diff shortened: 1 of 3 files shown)


src/test/java/com/kalessil/phpStorm/phpInspectionsEA/api/JsonEncodingApiUsageInspectorTest.java
```diff
@@ -38,4 +38,21 @@ public void testIfFindsErrorsHandlingPatterns() {
             myFixture.checkResultByFile("testData/fixtures/api/json-encoding-errors-handling.fixed.php");
         }
     }
+    public void testItRecognizesNameIdentifierFlags() {
+        final PhpLanguageLevel level = PhpLanguageLevel.parse("8.0");
+        if (level != null && level.getVersionString().equals("8.0")) {
```
ricardoboss, line 43: Shoudln't we test for everything greater than or equal to 8.0? (8.1, 8.2, etc)

winkbrace: Well, if it works in 8.0 it should also work in higher versions. I believe this will just generate an php version 8.0 PhpProjectConfigurationFacade

PR #1815, opened 2022-02-22
Improved detection of strict throws flag

ricardoboss:
This PR is an attempt to address #1725.

I'm not an experienced Java developer and certainly not a plugin dev. I was unable to run this locally due to some errors with `NotNull` annotations and `java.lang.Object` not being found...

I hope to get some support to properly run this locally and fix any issues on the way.

### DIFF
--- a/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/JsonEncodingApiUsageInspector.java
+++ b/src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/JsonEncodingApiUsageInspector.java
@@ -37,11 +37,12 @@ public class JsonEncodingApiUsageInspector extends BasePhpInspection {
 
     private static final String messageResultType     = "Please specify the second argument (clarifies decoding into array or object).";
     private static final String messageErrorsHandling = "Please consider taking advantage of JSON_THROW_ON_ERROR flag for this call options.";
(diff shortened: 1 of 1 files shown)


src/main/java/com/kalessil/phpStorm/phpInspectionsEA/inspectors/apiUsage/JsonEncodingApiUsageInspector.java
```diff
@@ -63,83 +64,121 @@ public PsiElementVisitor buildVisitor(@NotNull final ProblemsHolder holder, bool
...
+                                DECODE_AS_ARRAY ? "true" : "false"
+                        );
+                        holder.registerProblem(
+                                reference,
+                                MessagesPresentationUtil.prefixWithEa(messageResultType),
+                                DECODE_AS_ARRAY ? new DecodeIntoArrayFix(replacement) : new DecodeIntoObjectFix(replacement)
+                        );
+                    }
+
+                    if (HARDEN_ERRORS_HANDLING && arguments.length > 0 && PhpLanguageLevel.get(holder.getProject()).atLeast(PhpLanguageLevel.PHP730)) {
```
CRC-Mismatch, line 87: I didn't test it, but isn't named arguments a feature that was introduced only in PHP 8.0?

ricardoboss: Good catch!

PR #1842, opened 2022-08-03
Fix incorrect usage of "amount" instead of "number"

TravisCarden:
Many user messages incorrectly use the word "amount" where "number" belongs. See [Understanding The Difference Between Number and Amount](https://www.enago.com/academy/understanding-the-difference-between-number-and-amount/#:~:text=Although%20number%20and%20amount%20have,at%20the%20noun%20being%20described.). It's a niggle, but it distracts me a little every time I see it. 😛 It's just a string change, so hopefully it doesn't break any tests.


ea-inspections-team:
Thank you @TravisCarden, merged!

PR #1866, opened 2022-12-15
Improve docs for `Phar-incompatible 'realpath()' usage`

jdreesen:
You don't need two `dirname()` call because there's a second parameter for the level.


ea-inspections-team:
Thank you @jdreesen, merged!

